PR #48, opened 2023-05-11
Refresh caches on any SDK-driven creates or deletes

cmgrote:
- Forces a cache refresh any time a new custom metadata or classification typedef is created or deleted
- Renames `_refresh_cache` methods within the caches to `refresh_cache`, since they're no longer private


ErnestoLoma (approved):
Looks good to me

PR #49, opened 2023-05-12
fix: invalid default argument

gaczm:
Current default value for parameter `attributes` causes error:
>  none is not an allowed value (type=type_error.none.not_allowed)


gaczm:
Ok, thank you

PR #50, opened 2023-05-12
Adds shortcuts to create custom metadata properties

cmgrote:


cmgrote:
Actually, don't review / merge this yet — will incorporate the more extensive integration test, first.

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+            category=AtlanTypeCategory.ENUM,
+            name=name,
+            element_defs=EnumDef.ElementDef.list_from(values),
+        )
+
+    def get_valid_values(self) -> Optional[List[str]]:
+        """
+        Translate the element definitions in this enumeration into simple list of strings.
+        """
+        if self.element_defs:
```
ErnestoLoma, line 228: I would suggest replacing lines 228/234 with a list comprehension as follows:
`return [one.value for one in self.element_defs] if self.element_defs else []`

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+            AtlanCustomAttributePrimitiveType.USERS,
+            AtlanCustomAttributePrimitiveType.GROUPS,
+            AtlanCustomAttributePrimitiveType.URL,
+            AtlanCustomAttributePrimitiveType.SQL,
+        ):
+            base_type = AtlanCustomAttributePrimitiveType.STRING.value
+        else:
+            base_type = attribute_type.value
+        if multi_valued:
+            builder.type_name = "array<" + str(base_type) + ">"
```
ErnestoLoma, line 433: I would suggest using an f-string as follows:
`builder.type_name = f"array<{str(base_type)}>"`

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+            from pyatlan.cache.enum_cache import EnumCache
+
+            if enum_def := EnumCache.get_by_name(str(options_name)):
+                builder.enum_values = enum_def.get_valid_values()
+            else:
+                builder.enum_values = []
+        return builder
+
+    def is_archived(self) -> bool:
+        if opt := self.options:
```
ErnestoLoma, line 447: I would suggest replacing lines 447/450 with:
`return bool(opt.is_archived) if (opt := self.options) else False`

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+        else:
+            return False
+
+    def archive(self, by: str) -> AttributeDef:
+        if self.options:
+            removal_epoch = int(time.time() * 1000)
+            self.options.is_archived = True
+            self.options.archived_by = by
+            self.options.archived_at = removal_epoch
+            self.display_name = self.display_name + "-archived-" + str(removal_epoch)
```
ErnestoLoma, line 458: I would suggest to use of an f-string as follows:
`self.display_name = f"{self.display_name}-archived-{removal_epoch}"`

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+        def list_from(values: List[str]) -> List[EnumDef.ElementDef]:
+            from pyatlan.model.assets import validate_required_fields
+
+            validate_required_fields(
+                ["values"],
+                [values],
+            )
+            elements = []
+            if values:
+                for i in range(0, len(values), 1):
```
ErnestoLoma, line 196: I would suggest replacing 196/197 with the following:
`elements.extend(EnumDef.ElementDef.of(ordinal=i, value=values[i]) for i in range(len(values)))`

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+            options_name: Optional[str] = None,
+        ) -> AttributeDef.Options:
+            from pyatlan.model.assets import validate_required_fields
+
+            validate_required_fields(
+                ["type"],
+                [type],
+            )
+            # Explicitly set all defaults to ensure inclusion during pydantic serialization
+            builder = AttributeDef.Options(
```
ErnestoLoma, line 302: quible it's not a builder it's a AttributeDef.Options.

pyatlan/model/typedef.py
```diff
@@ -1,241 +1,462 @@
...
+        options_name: Optional[str] = None,
+    ) -> AttributeDef:
+        from pyatlan.model.assets import validate_required_fields
+
+        validate_required_fields(
+            ["display_name", "attribute_type"],
+            [display_name, attribute_type],
+        )
+        # Explicitly set all defaults to ensure inclusion during pydantic serialization
+        builder = AttributeDef(
```
ErnestoLoma, line 404: Again not a builder but an AttributeDef

tests/integration/classification_test.py
```diff
@@ -4,41 +4,50 @@
...
+from pyatlan.model.enums import AtlanClassificationColor, AtlanTypeCategory
 from pyatlan.model.typedef import ClassificationDef
 
-CLS_NAME = "psdk-ClassificationTest"
+CLS_NAME = "psdk-classification"
 
-# NOTE: Tests are intended to run in the order specified in this file,
-# to complete creation prior to testing prior to deletion.
 
+class ClassificationTest:
```
ErnestoLoma, line 13: If you add a class for called ClassificationTest then I would expect the tests to be part of the class not external to it.

tests/integration/classification_test.py
```diff
@@ -4,41 +4,50 @@
...
+    @staticmethod
+    def create_classification(client: AtlanClient, name: str) -> ClassificationDef:
```
ErnestoLoma, line 15: I'm not sure why this method is here. All the asserts in the method make it look like it's really a test.

tests/integration/classification_test.py
```diff
@@ -4,41 +4,50 @@
...
+        created = response.classification_defs[0]
+        assert created.category == AtlanTypeCategory.CLASSIFICATION
+        assert created.name
+        assert created.guid
+        assert created.name != name
+        assert created.display_name == name
+        return created
 
-@pytest.fixture
+    @staticmethod
```
ErnestoLoma, line 30: Again I don't understand why this is here as a staticmethod.

tests/integration/classification_test.py
```diff
@@ -4,41 +4,50 @@
...
-    cls = ClassificationDef(
-        name=CLS_NAME, display_name=CLS_NAME, options={"color": "GREEN"}
-    )
-    response = client.create_typedef(cls)
-    print(response)
-    assert response
-    assert response.classification_defs
-    assert len(response.classification_defs) == 1
+@pytest.fixture(scope="module")
+def cls(client: AtlanClient):
```
ErnestoLoma, line 41: Since this fixture returns a classification the convention would be to name is classification

tests/integration/custom_metadata_test.py
```diff
@@ -1,132 +1,855 @@
...
 
-def test_001_create_custom_metadata(client: AtlanClient):
-    attribute_defs = [
-        AttributeDef(
-            name=CM_ATTR_LICENSE,
-            display_name=CM_ATTR_LICENSE,
-            type_name="string",
-            options=AttributeDef.Options(),
+@pytest.fixture(scope="module")
+def cm_ipr(client: AtlanClient) -> Generator[CustomMetadataDef, None, None]:
```
ErnestoLoma, line 82: The presence of all the asserts makes this look like a test not a fixture. Because it's a fixture if any of the asserts before the yield fail then I don't believe that the code after the yield will be run.

tests/integration/custom_metadata_test.py
```diff
@@ -1,132 +1,855 @@
...
+    assert one.type_name == AtlanCustomAttributePrimitiveType.STRING.value
+    assert not one.options.multi_value_select
+    yield cm_created
+    # Cleanup after completion, after a little wait for eventual consistency
+    time.sleep(5)
+    client.purge_typedef(cm_created.name)
+
+
+@pytest.fixture(scope="module")
+def cm_enum(client: AtlanClient) -> Generator[EnumDef, None, None]:
```
ErnestoLoma, line 237: Again this looks like a test not a fixture

ErnestoLoma (changes requested):
I sent you a meeting request so we can discuss how we might have a more pythonic integration tests. Let me know if there's a better time for you.

tests/integration/classification_test.py
cmgrote: Actually returns a `classification_def`, so will name it that instead (and also add the missing return type hint 🙈)

ErnestoLoma (approved):
Looks good.

ErnestoLoma (approved):
Looks good.

ErnestoLoma (approved):
Looks good.

PR #61, opened 2023-06-05
Refactor usage of custom metadata

ErnestoLoma:
Refactored the usage of custom metadata from property based access to properties of metadata sets to string indexed acccess.


cmgrote (approved):
Looks good to me

PR #63, opened 2023-06-08
Adds icons and images for classifications

cmgrote:


pyatlan/client/atlan.py
```diff
@@ -341,6 +344,66 @@ def _call_api(
                     )
             raise AtlanServiceException(api, response)
 
+    def _upload_file(self, api, file=None, filename=None):
+        generator = MultipartDataGenerator()
+        generator.add_file(file=file, filename=filename)
+        post_data = generator.get_post_data()
+        api.produces = "multipart/form-data; boundary=%s" % (generator.boundary,)
```
ErnestoLoma, line 351: Please use f-string:
`api.produces = f"multipart/form-data; boundary={generator.boundary}"`

pyatlan/client/atlan.py
```diff
@@ -341,6 +344,66 @@ def _call_api(
...
+                error_info = json.loads(response.text)
+                error_code = error_info.get("errorCode", 0)
+                error_message = error_info.get("errorMessage", "")
+                if error_code and error_message:
+                    raise AtlanError(
+                        message=error_message,
+                        code=error_code,
+                        status_code=response.status_code,
+                    )
+            raise AtlanServiceException(api, response)
```
ErnestoLoma, line 405: This looks like a duplicate of lines 298/345 with the exception of filename vs request_obj on line 375. Please refactor to a method.

pyatlan/multipart_data_generator.py
```diff
@@ -0,0 +1,79 @@
...
+            self._write("Content-Type: image/webp")
+        else:
+            self._write("Content-Type: application/octet-stream")
+        self._write(self.line_break)
+        self._write(self.line_break)
+        self._write_file(file)
+        self._write(self.line_break)
+
+    def param_header(self):
+        return "--%s" % self.boundary
```
ErnestoLoma, line 52: Please use f-string:
`return f"--{self.boundary}"`

pyatlan/multipart_data_generator.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def get_post_data(self):
+        self._write("--%s--" % (self.boundary,))
```
ErnestoLoma, line 55: Please use f-string:
`self._write(f"--{self.boundary}--")`

pyatlan/multipart_data_generator.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+
+        self.data.write(array)
+
+    def _write_file(self, f):
+        while True:
+            file_contents = f.read(self.chunk_size)
+            if not file_contents:
+                break
+            self._write(file_contents)
```
ErnestoLoma, line 76: I think the following is a bit better:
```
if file_contents := f.read(self.chunk_size):
     self._write(file_contents)
 else:
      break
```

pyatlan/multipart_data_generator.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _initialize_boundary(self):
```
ErnestoLoma, line 78: This doesn't need to be a method. It could just be a function in the module or a static method.

tests/integration/classification_test.py
```diff
@@ -54,36 +57,97 @@ def _make_classification(name: str) -> ClassificationDef:
...
+    client: AtlanClient,
+    image: AtlanImage,
+    make_unique: Callable[[str], str],
+) -> Generator[ClassificationDef, None, None]:
+    cls_name = f"{make_unique(MODULE_NAME)}_image"
+    cls = ClassificationDef.create(
+        name=cls_name, color=AtlanClassificationColor.YELLOW, image=image
+    )
+    response = client.create_typedef(cls)
+    created = response.classification_defs[0]
```
ErnestoLoma, line 85: No need for a variable. Just yield immediately.

ErnestoLoma (changes requested):
I would also like to see a unit test for multipart_data_generator, if possible.

pyatlan/multipart_data_generator.py
cmgrote: Thinking this is even better (while True with a break clause just feels wrong to me):

```
        while file_contents := f.read(self.chunk_size):
            self._write(file_contents)
```

ErnestoLoma (approved):

PR #64, opened 2023-06-08
Migrates to easier to use uniqueness generator for testing

cmgrote:


cmgrote:
One other change I made here — backed-off the triggering of GitHub Actions so that they'll only run as part of a pull request (including any new code pushes to a PR). I don't see a need to run them against every single code push (outside of PRs) — that'll just increase the amount of parallelism and integration test environment overflow, without any added benefit (we don't merge anything into `main` other than via a PR anyway, since `main` is a protected branch)

ErnestoLoma (approved):

PR #69, opened 2023-06-14
Fix problem with Badge creation

ErnestoLoma:
Explicitly set status to Active when in create method of Badge.


cmgrote (changes requested):
Please avoid regenerating typedefs for the moment as I've polluted the tenant with some test objects (temporarily).

cmgrote (approved):

PR #70, opened 2023-06-15
ACTIV-670 ACTIV-710

ErnestoLoma:
Added trim_to_required
Added create to Asset which raises Unimplemented exception
Set the view_name/table_name in the Column.create.


pyatlan/generator/templates/entity.jinja2
```diff
@@ -907,6 +921,28 @@ class {{ entity_def.name }}({{super_classes[0]}} {%- if "Asset" in super_classes
                 name=name, anchor=anchor, parent_category=parent_category
             )
         )
+
+    def trim_to_required(self) -> AtlasGlossaryCategory:
+        return self.create_for_modification(qualified_name=self.qualified_name, name=self.name, glossary_guid = self.anchor.guid)
```
cmgrote, line 926: What happens if `self.anchor` is `None` (hasn't yet been set to anything)? Won't `self.anchor.guid` throw an error then? If so, we should add the additional checks here to avoid that error, as it probably won't be obvious if we just allow that error to be thrown (?)

pyatlan/generator/templates/entity.jinja2
```diff
@@ -925,6 +961,8 @@ class {{ entity_def.name }}({{super_classes[0]}} {%- if "Asset" in super_classes
                 name=name, anchor=anchor, glossary_qualified_name=glossary_qualified_name,  glossary_guid = glossary_guid, categories=categories
             )
         )
+    def trim_to_required(self) -> AtlasGlossaryTerm:
+        return self.create_for_modification(qualified_name=self.qualified_name, name=self.name, glossary_guid = self.anchor.guid)
```
cmgrote, line 965: As above — extra check for `self.anchor` before trying to resolve `self.anchor.guid` (?)

cmgrote (commented):
Not sure if we need an extra validation for `anchor` before trying to do `anchor.guid` (?)

pyatlan/generator/templates/entity.jinja2
ErnestoLoma: Done.

cmgrote:
I don't see any revision pushed after the review — have you pushed your commits with the updates?

cmgrote (approved):

PR #75, opened 2023-06-22
Activ 486

ErnestoLoma:
Rename all references to Classification and classification to Tag and tag (TagDef, Tag, tags, tagNames, etc):


pyatlan/generator/templates/methods/attribute/purpose.jinja2
```diff
@@ -1,13 +1,13 @@
...
+        def create(cls, name: str, atlan_tags: list[str]) -> {{ entity_def.name }}.Attributes:
+            validate_required_fields(["name", "atlan_tags"], [name, atlan_tags])
             return Purpose.Attributes(
                 qualified_name=name,
                 name=name,
                 display_name=name,
                 is_access_control_enabled=True,
                 description="",
-                purpose_classifications=classifications
+                purpose_classifications=atlan_tags
```
cmgrote, line 12: Can we alias `purpose_classifications` here, too?

pyatlan/generator/templates/referenceable_attributes.jinja2
```diff
@@ -72,8 +72,8 @@
     version: Optional[int] = Field(
         None, description='Version of this object.\n', example=2
     )
-    classifications: Optional[list[Classification]] = Field(
-        None, description="classifications"
+    atlan_tags: Optional[list[AtlanTag]] = Field(
+        None, description="classifications", alias="classifications"
```
cmgrote, line 76: I would change the description to "Atlan tags" as well — we should only have `classification(s)` appear as an alias, and literally nowhere else (if possible)

tests/integration/purpose_test.py
```diff
@@ -24,22 +24,21 @@
 @pytest.fixture(scope="module")
 def atlan_tag(
     client: AtlanClient,
-) -> Generator[ClassificationDef, None, None]:
-    classification_def = ClassificationDef.create(
+) -> Generator[AtlanTagDef, None, None]:
+    classification_def = AtlanTagDef.create(
```
cmgrote, line 28: I'd:
1. Change the variable name to `atlan_tag_def` for consistency
2. Change the name of the enum from `AtlanClassificationColor` to `AtlanTagColor`

cmgrote (changes requested):
Also, in the `Tag` typedef there is an attribute called `mappedClassificationName` which I think we need to rename to `mappedAtlanTagName` (and alias back to the original name)

tests/integration/purpose_test.py
ErnestoLoma: done

pyatlan/generator/templates/referenceable_attributes.jinja2
ErnestoLoma: done

cmgrote (approved):

PR #76, opened 2023-06-22
Adds file management

cmgrote:


ErnestoLoma (changes requested):
Could you add unit tests similar to those for other assets in the tests/unit/model package.

ErnestoLoma (approved):

PR #77, opened 2023-06-23
Event-handling abstractions

cmgrote:


pyatlan/events/atlan_event_handler.py
```diff
@@ -0,0 +1,151 @@
...
+    return WEBHOOK_VALIDATION_REQUEST == data
+
+
+def valid_signature(expected: str, headers: dict[str, str]) -> bool:
+    """
+    Validate the signing secret provided with a request matches the expected signing secret.
+    """
+    if not headers:
+        return False
+    else:
```
ErnestoLoma, line 23: The else is not needed

pyatlan/events/atlan_event_handler.py
```diff
@@ -0,0 +1,151 @@
...
+        dsl=dsl,
+        attributes=attributes,
+        relation_attributes=["guid"],
+        exclude_meanings=~include_meanings,
+        exclude_atlan_tags=~include_atlan_tags,
+    )
+    response = client.search(criteria=request)
+    if result := response.current_page()[0]:
+        return result
+    return None
```
ErnestoLoma, line 61: A bit more succinct:
``return result if (result := response.current_page()[0]) else None``

pyatlan/events/atlan_event_handler.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def has_description(asset: Asset) -> bool:
+    """
+    Check if the asset has either a user-provided or system-provided description.
+    """
+    description = asset.user_description
+    if not description:
+        description = asset.description
+    return description is not None and description != ""
```
ErnestoLoma, line 71: A bit more succinct:
`description = asset.user_description or asset.description`
`return description is not None and description != ""`

pyatlan/events/atlan_event_handler.py
```diff
@@ -0,0 +1,151 @@
...
+    # If possible, look directly on inputs and outputs rather than the __hasLineage flag
+    if isinstance(asset, Catalog):
+        return (asset.input_to_processes is not None) or (
+            asset.output_from_processes is not None
+        )
+    else:
+        return bool(asset.has_lineage)
+
+
+class AtlanEventHandler:
```
ErnestoLoma, line 94: Is this intended to be subclassed and not used directly. If so it should extend [ABC](https://docs.python.org/3/library/abc.html)

pyatlan/events/atlan_lambda_handler.py
```diff
@@ -0,0 +1,38 @@
...
+SIGNING_SECRET = str(os.environ.get("SIGNING_SECRET"))
+
+
+def process_event(handler: AtlanEventHandler, event, context):
+    body = event.get("body")
+    if is_validation_request(body):
+        print("Matches a validation request - doing nothing and succeeding.")
+        return {"statusCode": 200}
+    else:
+        if valid_signature(SIGNING_SECRET, event.get("headers")):
```
ErnestoLoma, line 23: It changed the if to if not and move up raising the exception. Then the else would not be needed:
``` if not valid_signature(SIGNING_SECRET, event.get("headers")):```
```    raise IOError("Invalid signing secret received - will not process this request.")```

File: tests/unit/test_events.py [hunk 6]
```diff
@@ -0,0 +1,116 @@
...
+    '\\"bacab52b-6c4b-4dbe-b5da-97ec2e509c7e\\",\\"status\\":\\"ACTIVE\\",\\"displayText\\":\\"Example\\"},'
+    '\\"operationType\\":\\"ENTITY_CREATE\\",\\"eventTime\\":1666792332986}}}",'
+    '  "isBase64Encoded": false'
+    "}"
+)
+
+
+def test_validation_payload():
+    event = json.loads(VALIDATION_PAYLOAD)
+    body = event.get("body")
```
ErnestoLoma, line 105: I don't understand why you are bothering with loading the JSON since the test is just use a portion of it. You could just as well have just passed the requisite dict.

tests/unit/test_events.py
```diff
@@ -0,0 +1,116 @@
...
+    assert is_validation_request(body)
+
+
+def test_no_signing_key():
+    event = json.loads(VALIDATION_PAYLOAD)
```
ErnestoLoma, line 110: Same as above

tests/unit/test_events.py
```diff
@@ -0,0 +1,116 @@
...
+    assert not valid_signature(SIGNING, event.get("headers"))
+
+
+def test_signing_key():
```
ErnestoLoma, line 114: Same as above

ErnestoLoma (changes requested):
I would have expected some unit tests to instantiate the events from JSON. Is the too difficult? Am I missing something.

ErnestoLoma (approved):

PR #79, opened 2023-06-28
Activ 737 - Port findByName methods for glossary-related assets

ErnestoLoma:


pyatlan/model/search.py
```diff
@@ -370,7 +371,9 @@ def with_created_by(cls, value: StrictStr):
 
     @classmethod
     @validate_arguments()
-    def with_glossary(cls, value: StrictStr):
+    def with_glossary(
+        cls, value: constr(strip_whitespace=True, min_length=1, strict=True)  # type: ignore
```
cmgrote, line 375: Should we rename the `value` argument to `qualified_name` (or `name`)? To make it clear to someone using the method which input they're expected to provide (`value` is too generic, I think?)

cmgrote (changes requested):

pyatlan/model/search.py
ErnestoLoma: Done.

tests/integration/glossary_test.py
```diff
@@ -49,6 +56,28 @@ def test_glossary(
...
+    assert category.guid
+    assert category.name == MODULE_NAME
+    assert category.qualified_name
+    c = client.get_asset_by_guid(category.guid, AtlasGlossaryCategory)
+    assert c
+    assert c.guid == category.guid
+    assert c.anchor
+    assert c.anchor.guid == glossary.guid
+
+
```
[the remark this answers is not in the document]
ErnestoLoma: I moved the tests to this class.

cmgrote (approved):

PR #80, opened 2023-06-29
Adds creation and upload of a zip file for Lambda usage with each release

cmgrote:


ErnestoLoma (approved):
I don't totally understand this. Looks like you are installing pyatlan in a directory called python and then creating a zip file and uploading to our github repo.

PR #82, opened 2023-06-30
Change lineage list response into a lazy-fetching generator

cmgrote:


pyatlan/client/atlan.py
```diff
@@ -287,6 +282,56 @@ def __iter__(self) -> Generator[Asset, None, None]:
...
+            assets: list[Asset],
+        ):
+            self._client = client
+            self._criteria = criteria
+            self._start = start
+            self._size = size
+            self._has_more = has_more
+            self._assets = assets
+
+        def has_more(self) -> bool:
```
ErnestoLoma, line 302: Add the @property annotation to make this a property

pyatlan/client/atlan.py
```diff
@@ -287,6 +282,56 @@ def __iter__(self) -> Generator[Asset, None, None]:
                 if not self.next_page():
                     break
 
+    class LineageResults:
```
ErnestoLoma, line 285: There's a lot of duplicate code between LineageResults and SearchResults. So I think maybe the following refactoring would be in good:

1. Create an abstract class SearchRequest which both LineageListRequest and IndexSearchRequest extend.
2. Change SearchResults to take a criteria that is a SearchRequest
3. Create abstract properties in SearchRequest for:
    1. has_more
    2. count
    4. offset
    5. size

ErnestoLoma (changes requested):
I suggested a bit of refactoring

pyatlan/client/atlan.py
cmgrote: I think only `offset` and `size` are shared:
- Only lineage list has the `has_more` (does not have a `count`)
- Only a indexsearch has a `count` (but does not have a `has_more`)

ErnestoLoma: That's why I suggest added different implementations to the sub classes that way we can eliminate the duplication.

ErnestoLoma (approved):

PR #88, opened 2023-07-12
Refactoring to enable per-client caching

cmgrote:
Most (all?) of the integration tests should be working; however, various unit tests are now broken — given the need now for a client to be in place for the caches to be usable. (Not sure if resolving those is a simple fix / monkeypatch?)

Also not sure how we want to handle the setup of the `CustomMetadataProxy` embedded in `Referenceable`, as this would imply we need to initialize any / every object with a client to have a CustomMetadataCache available for building up that proxy (?)


pyatlan/generator/templates/methods/asset/connection.jinja2
```diff
@@ -9,30 +9,30 @@
         admin_users: Optional[list[str]] = None,
         admin_groups: Optional[list[str]] = None,
         admin_roles: Optional[list[str]] = None,
+        client: Optional[AtlanClient] = None,
```
ErnestoLoma, line 12: If we are going to add a client as a parameter then I think it should be in Referenceable, but I don't really like the need this dependence on a client.

pyatlan/model/core.py
```diff
@@ -60,10 +54,13 @@ def to_snake_case(value):
 
 
 class AtlanTagName:
-    def __init__(self, display_text: str):
-        from pyatlan.cache.atlan_tag_cache import AtlanTagCache
-
-        if not AtlanTagCache.get_id_for_name(display_text):
+    def __init__(
+        self,
+        display_text: str,
```
ErnestoLoma, line 59: Add ability to pass client otherwise you need to remember to set the default_client correctly.

pyatlan/model/core.py
```diff
@@ -87,21 +84,26 @@ def __eq__(self, other):
         )
 
     @classmethod
-    def _convert_to_display_text(cls, data):
-        from pyatlan.cache.atlan_tag_cache import AtlanTagCache
-
+    def _convert_to_display_text(
+        cls,
+        data,
```
ErnestoLoma, line 89: Need to add client parameter otherwise you need to remember to set the correct default_client

pyatlan/model/core.py
```diff
@@ -87,21 +84,26 @@ def __eq__(self, other):
...
             return AtlanTagName(display_text)
         else:
             raise ValueError(f"{data} is not a valid AtlanTag")
 
     @staticmethod
-    def json_encode_atlan_tag(atlan_tag_name: "AtlanTagName"):
-        from pyatlan.cache.atlan_tag_cache import AtlanTagCache
-
-        return AtlanTagCache.get_id_for_name(atlan_tag_name._display_text)
+    def json_encode_atlan_tag(
```
ErnestoLoma, line 101: Same comment regarding need for client as parameter

ErnestoLoma (changes requested):
I think the fundamental problem is the usage of a class variable for the client. This means there can only be one value for all instances in the current run. I don't  adding  AtlanClient.get_default_client_or_fail  to the code without the opportunity to inject an instance of the client improves anything. You must still remember to reset the default client.  
In pyatlan.client.atlan you changed all occurences of list to List. I don't think this change was necessary. 
Also a number of places you commented out the @validate_arguments annotation. I'm not sure why this was required. Please add a comment explaining why it is commented out so it doesn't  get added back

pyatlan/client/atlan.py
```diff
@@ -219,27 +138,445 @@ def _build_typedef_request(typedef: TypeDef) -> TypeDefResponse:
     return payload
 
 
-def _refresh_caches(typedef: TypeDef) -> None:
-    if isinstance(typedef, AtlanTagDef):
-        from pyatlan.cache.atlan_tag_cache import AtlanTagCache
+class AtlanClient(BaseSettings):
```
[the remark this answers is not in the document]
cmgrote: 1. They're managed and maintained as part of a client now
2. Avoids circular references between `AtlanClient` (containing cache instances) and `...Cache` (which requires an `AtlanClient` instance in order to refresh its contents)

pyatlan/client/atlan.py
```diff
@@ -1172,11 +1539,11 @@ def remove_custom_metadata(self, guid: str, cm_name: str):
             custom_metadata_request,
         )
 
-    @validate_arguments()
+    # @validate_arguments()
```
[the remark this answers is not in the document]
cmgrote: Basically `validate_arguments()` doesn't work with forward references — https://github.com/pydantic/pydantic/issues/2901

pyatlan/client/atlan.py
```diff
@@ -1,14 +1,16 @@
...
+
 import abc
 import contextlib
 import copy
 import json
 import logging
 import os
 from abc import ABC
-from typing import ClassVar, Generator, Optional, Type, TypeVar, Union
+from typing import ClassVar, Generator, List, Optional, Type, TypeVar, Union
```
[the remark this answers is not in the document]
cmgrote: You might be right, this could be left-over from using `ForwardRef()` to try to workaround `validate_arguments()` lack of support for forward references (which didn't end up working either, anyway, and seems to be an anti-pattern in Python despite being suggested in Pydantic's own docs)

Ref:
- Python docs (note from PEP 585 you must use `List` and not `list`): https://docs.python.org/3.11/library/typing.html#typing.ForwardRef
- Pydantic's docs: https://docs.pydantic.dev/1.10/usage/postponed_annotations/

pyatlan/cache/custom_metadata_cache.py
```diff
@@ -1,10 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2022 Atlan Pte. Ltd.
-from typing import Optional
-
-from pyatlan.error import InvalidRequestError, LogicError, NotFoundError
-from pyatlan.model.enums import AtlanTypeCategory
-from pyatlan.model.typedef import AttributeDef, CustomMetadataDef
 
 
 class Synonym:
```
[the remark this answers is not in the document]
cmgrote: Yeah, from what I can see this was your code — happy to remove it if you're confirm it's not used 👍

pyatlan/generator/templates/methods/asset/connection.jinja2
cmgrote: I don't either — for me it comes down to when we do the validation:

- At the time of object creation (advantage is we fail fast)
- At the time of object marshalling (just before sending to Atlan) — would probably be the more logical point to inject the client, but means any potential failure is delayed

pyatlan/client/atlan.py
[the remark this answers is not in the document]
cmgrote: Yup, looks like it's unnecessary with the current approach — have reverted in latest commit I pushed

pyatlan/client/atlan.py
```diff
@@ -1472,6 +1839,104 @@ def find_term_by_name(
...
+from pyatlan.model.user import (  # noqa: E402
+    AddToGroupsRequest,
+    AtlanUser,
+    ChangeRoleRequest,
+    CreateUserRequest,
+    UserMinimalResponse,
+    UserResponse,
+)
+
+# Assets = Union[
```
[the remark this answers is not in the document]
cmgrote: To test just that — IDE couldn't find anywhere it was used, but I don't actually know what it was intended to do, so was testing accordingly. Guess this confirms it's not used?

pyatlan/client/atlan.py
```diff
@@ -356,6 +693,12 @@ def register_client(cls, client: "AtlanClient"):
     def get_default_client(cls) -> "Optional[AtlanClient]":
         return cls._default_client
 
+    @classmethod
+    def get_default_client_or_fail(cls) -> "AtlanClient":
```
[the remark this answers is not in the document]
ErnestoLoma: I think this is preferable. Now if I want to check if a default client has been set I have to put it in a try block.

pyatlan/model/core.py
cmgrote: Can add, but unclear to me how this would ultimately be passed through via `__get_validators__` (which if it's always sending as the default `None` then would behave identically to not passing it through at all, I think?)

pyatlan/model/core.py
cmgrote: Same comment above — I don't know how / if this will ever be sent a value by Pydantic, or will just always default to `None` (?)

pyatlan/model/core.py
cmgrote: And indeed, Pydantic throws an error if we pass through a client parameter:

```
pydantic.errors.ConfigError: Invalid signature for validator <bound method AtlanTagName._convert_to_display_text of <class 'pyatlan.model.core.AtlanTagName'>>: (data, client: 'Optional[AtlanClient]' = None), should be: (value, values, config, field), "values", "config" and "field" are all optional.
```

PR #97, opened 2023-08-10
Adds further client methods

cmgrote:
- Bulk delete
- Restore
- Search by certificate
- Activate / deactivate users
- Search by any of multiple types

+ regenerates model to include latest assets and structs


cmgrote:
Just fixing all the broken unit tests (we should probably generate these longer-term)...

PR #98, opened 2023-08-10
Enhancements + removes user deletion functionality

cmgrote:
And adds functionality:
- Bulk delete
- Restore
- Search by certificate
- Activate / deactivate users
- Search by any of multiple types

(Also regenerates model to include latest assets and structs)

### DIFF
--- a/tests/integration/s3_asset_test.py
+++ b/tests/integration/s3_asset_test.py
@@ -58,9 +58,8 @@ def _retrieve_and_check(client: AtlanClient, to_check: list[Asset], retry_count:
     if leftovers:
         if retry_count == 20:
             raise AtlanError(message="Overran retry limit", code="500", status_code=500)
-        else:
-            time.sleep(2)
-            _retrieve_and_check(client, leftovers, retry_count + 1)
+        time.sleep(2)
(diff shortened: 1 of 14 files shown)


tests/integration/s3_asset_test.py
```diff
@@ -16,6 +25,42 @@
...
+        except NotFoundError:
+            # If it is not found, it was successfully deleted (purged), so we
+            # do not need to look for it any further
+            print("Asset no longer exists.")
+        except AtlanError:
+            leftovers.append(one)
+    if leftovers:
+        if retry_count == 20:
+            raise AtlanError(message="Overran retry limit", code="500", status_code=500)
+        else:
```
ErnestoLoma, line 61: This else isn't needed since an exception is being raised.

ErnestoLoma (approved):
Minor change requested.

tests/integration/s3_asset_test.py
ErnestoLoma: The if retry_count == 20 is needed to check if the exception should be raised. The else isn't needed because when the exception is raised the code block will be exited.

cmgrote: Ooohhh, I see what you mean now 😁

PR #102, opened 2023-08-15
Enhancements and fixes

cmgrote:
- Adds new purpose policy permissions (attaching / detaching terms from assets)
- Fixes type for `decentralized_role`
- Breaking: `find_category_by_name[_fast]` now return a list of categories rather than a single category, since categories are not unique by name within a glossary.
- Documents various public methods in client, caches, and event-handling classes

### DIFF
--- a/pyatlan/events/atlan_event_handler.py
+++ b/pyatlan/events/atlan_event_handler.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2023 Atlan Pte. Ltd.
 from abc import ABC
-from typing import Iterable, List, Optional
+from typing import Iterable, Optional
 
 from pyatlan.client.atlan import AtlanClient
 from pyatlan.model.assets import Asset, Catalog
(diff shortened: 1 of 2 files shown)


pyatlan/model/group.py
```diff
@@ -42,7 +42,7 @@ class Attributes(AtlanObject):
     attributes: Optional[AtlanGroup.Attributes] = Field(
         description="Detailed attributes of the group."
     )
-    decentralized_roles: Optional[str] = Field(description="TBC")
+    decentralized_roles: Optional[List[Any]] = Field(description="TBC")
```
ErnestoLoma, line 45: annotation should be Optional[list[Any]]

pyatlan/events/atlan_event_handler.py
```diff
@@ -143,12 +173,30 @@ def has_changes(self, current: Asset, modified: Asset) -> bool:
...
+        :param current: the current view / state of the asset in Atlan, that was the starting point for any change
+                        calculations
+        :param modified: the in-memory-modified asset against which to check if any changes actually need to be sent
+                         to Atlan
+        :returns: True if the modified asset should be sent on to (updated in) Atlan, or False if there are no actual
+                  changes to apply
         """
         return current == modified
 
     def upsert_changes(self, changed_assets: List[Asset]):
```
ErnestoLoma, line 186: Annotation should be list[Asset]

ErnestoLoma (changes requested):
A few cosmetic changes requested

PR #103, opened 2023-08-17
Refactor code generation

ErnestoLoma:
Refactored code generation to generate enums from enum-defs and split assets into multiple modules.


pyatlan/generator/templates/module.jinja2
```diff
@@ -2,10 +2,26 @@
 # Copyright 2022 Atlan Pte. Ltd.
 # Based on original code from https://github.com/apache/atlas (under Apache-2.0 license)
```
cmgrote, line 3: Is this still a true statement for the generated model files? I wouldn't think there is much of anything we've reused from Atlas to do this strongly-typed implementation of the models (?)

cmgrote (approved):

pyatlan/generator/templates/module.jinja2
ErnestoLoma: I've removed it

PR #104, opened 2023-08-22
Search simplifications and optimizations

cmgrote:
- Adds a new FluentSearch approach to searching (no need to know Elastic anymore)
- Changes default inclusion operation for Bool queries to `filter` rather than `must` (for efficiency)


pyatlan/model/fields/atlan_fields.py
```diff
@@ -0,0 +1,528 @@
...
+    atlan_field_name: StrictStr
+
+
+class RelationField(AtlanField):
+    """
+    Represents any field used to capture a relationship in Atlan, which is not inherently
+    searchable.
+    """
+
+    def __init__(self, atlan: StrictStr):
```
ErnestoLoma, line 36: The parameter should be called atlan_field_name not atlan. Same for the sub-classes

pyatlan/model/fields/atlan_fields.py
```diff
@@ -0,0 +1,528 @@
...
+        """
+        Default constructor.
+
+        :param atlan: name of the attribute in the metastore
+        :param elastic: name of the field in the search index
+        """
+        self.atlan_field_name = atlan
+        self.elastic_field_name = elastic
+
+    def exists(self) -> Query:
```
ErnestoLoma, line 62: I would make this and similar read only methods properties.

pyatlan/model/fluent_search.py
```diff
@@ -0,0 +1,391 @@
...
+        where_nots: Optional[list[Query]] = None,
+        where_somes: Optional[list[Query]] = None,
+        _min_somes: int = 1,
+    ):
+        self.wheres = wheres
+        self.where_nots = where_nots
+        self.where_somes = where_somes
+        self._min_somes = _min_somes
+
+    def _clone(self) -> "CompoundQuery":
```
ErnestoLoma, line 145: The convention in python is to override the __copy__ and __deepcopy__ methods. See [copy- Shallow and deep copy operations](https://docs.python.org/3/library/copy.html) and [Shallow and Deep Copy in Python and How to Use __copy__](https://thepythoncodingbook.com/2022/08/10/shallow-and-deep-copy-in-python/) Maybe just add a TODO comment to refactor in the future.

pyatlan/model/fields/atlan_fields.py
cmgrote: `exists` is literally the only one that has no parameters — every single other method has at least 1 parameter. I think it would be odd if this one way to get a `Query` was a property while _all_ the others are methods (?)

cmgrote:
Can you also double-check the Jinja template I removed is in fact no longer used (`assets/referenceable.jinja2`)? It looked to me like it's been replaced by other files (I couldn't find any reference to it in the generator code, either)...

pyatlan/model/fields/atlan_fields.py
ErnestoLoma: In this class yes, but in all the subclasses you have made the methods that don't have parameters properties so I think it would be inconsistent for this to be a method while all the others are properties in the sub-classes.

ErnestoLoma:
> 

Yes it has been replaced by templates/referenceable_attributes.jinja.

ErnestoLoma (changes requested):

ErnestoLoma (approved):

PR #109, opened 2023-09-05
Am 176

ErnestoLoma:
Add retry for connnection creation and soft delete


pyatlan/client/atlan.py
```diff
@@ -173,18 +174,27 @@
     Type[View],
     Type[MaterialisedView],
 ]
+HTTPS_PREFIX = "https://"
+
+DEFAULT_RETRY = Retry(
+    total=4,
```
cmgrote, line 180: Let's drop default retries back to `3` and add `403` back into the list

cmgrote (approved):

PR #113, opened 2023-09-07
Am 146

ErnestoLoma:
Refactoring handling of exceptions


cmgrote (approved):
Looks very nice! Only additional thing that would be great for us to do still is to scan for any `raise ` statements, and ensure that we migrate any that are currently directly doing a `raise ValueError` or the like move over to the new `ErrorCode` enum-based approach.

(Fine for me if we do that via a separate PR, though.)

PR #126, opened 2023-09-20
Adds an API token check as part of cache refreshes

cmgrote:


pyatlan/cache/atlan_tag_cache.py
```diff
@@ -64,8 +65,10 @@ def _refresh_cache(self) -> None:
         Refreshes the cache of Atlan tags by requesting the full set of Atlan tags from Atlan.
         """
         response = self.provider.get_typedefs(
-            type_category=AtlanTypeCategory.CLASSIFICATION
+            type_category=[AtlanTypeCategory.CLASSIFICATION, AtlanTypeCategory.STRUCT]
```
ErnestoLoma, line 68: Why not raise the exception in the AtlanClient if all the lists are empty? That would remove some duplication of code.

pyatlan/cache/custom_metadata_cache.py
```diff
@@ -178,8 +178,10 @@ def _refresh_cache(self) -> None:
         :raises LogicError: if duplicate custom attributes are detected
         """
         response = self.provider.get_typedefs(
-            type_category=AtlanTypeCategory.CUSTOM_METADATA
+            type_category=[AtlanTypeCategory.CUSTOM_METADATA, AtlanTypeCategory.STRUCT]
         )
+        if not response or not response.struct_defs:
```
ErnestoLoma, line 183: Same comment as above

ErnestoLoma (changes requested):

pyatlan/cache/atlan_tag_cache.py
cmgrote: Each cache is only retrieving the minimal info it requires — so if some code _only_ needs to deal with tags, there's no need to retrieve all the custom metadata definitions (and vice versa). If we moved this up to the client level, then we'd need to retrieve the entire set of all type definitions, irrespective of whether someone is actually going to make use of them in their code or not (can add significant, unnecessary delay).

ErnestoLoma: I'm not suggesting retrieving all the metadata but if all the lists are empty, which I believe is what we generally see, that's a problem.

PR #128, opened 2023-09-22
Mark lineage method (and others) deprecated, including DeprecationWarning

cmgrote:


ErnestoLoma (approved):
Nice. I didn't know about this.

PR #129, opened 2023-09-26
Am 257 - Calculate aggregations

ErnestoLoma:


pyatlan/client/atlan.py
```diff
@@ -278,13 +279,15 @@ def __init__(
             start: int,
             size: int,
             assets: list[Asset],
+            aggregations: Optional[Aggregations],
```
cmgrote, line 282: I'm not sure this should be at the ABC-level — only index search will allow aggregations (not lineage listing, which also derives from `SearchResults` iirc), and afaik aggregations themselves are not paged — so no real benefit to having them in the ABC, I think?

pyatlan/client/atlan.py
```diff
@@ -2023,13 +2050,15 @@ def get_lineage_list(
         else:
             assets = []
             has_more = False
+        aggregations = self.get_aggregations(raw_json)
```
cmgrote, line 2053: There are no aggregations via the lineage list endpoint, they should only be used via index search.

pyatlan/model/fluent_search.py
```diff
@@ -350,8 +374,8 @@ def to_request(self) -> IndexSearchRequest:
             dsl.size = self._page_size
         if self.sorts:
             dsl.sort = self.sorts
-        # if self.aggregations:
-        #     dsl.aggregations.extend(self.aggregations)
+        if self.aggregations:
+            dsl.aggregations = self.aggregations
```
cmgrote, line 378: While small, doesn't the `=` here rather than an `extend` risk that we could clobber some aggregations if for some reason someone's managed to put aggregations in at different places?

cmgrote (changes requested):

cmgrote (approved):

cmgrote (approved):

PR #135, opened 2023-09-28
AM-247 : Manage ADLS assets

0xquark:
Please let me know if there are any tips or best practices I can follow to make my code better!

I've made changes to support ADLS assets in a way similar to how we manage AWS S3 assets.


cmgrote:
Haven't had a chance to scan through everything, but at a quick glance things look pretty good! I will review further tomorrow...

@ErnestoLoma if you have a chance, can you take a peek at the test failures in the log and comment back on the likely problem / fix? 🙏  Hoping it's something you've seen in the past and would have quick pointers on, otherwise I'll have a look myself tomorrow when I scan through the rest. 😁

cmgrote:
Ok, I have a hunch on the test failures — I think it's because the build may actually be running against your fork, which doesn't have the necessary secrets configured to access the environment. Just checking this hypothesis in a separate PR...

cmgrote (commented):
Only one thing that looked a bit odd to me, rest looks great 👍

0xquark:
Closing this in reference to #137

PR #140, opened 2023-10-05
Allow empty values for logos, equivalent to no logo

cmgrote:


cmgrote:
Very simple change, but just want an extra pair of 👀 to confirm having an enum value of `""` won't cause any strange problems (?)

ErnestoLoma (commented):
Looks good

ErnestoLoma (approved):
Looks good

PR #141, opened 2023-10-05
AM-220   Search activity log / asset history

ErnestoLoma:


pyatlan/model/custom_metadata.py
```diff
@@ -4,33 +4,59 @@
...
+
+    _sentinel: Optional["CustomMetadataDict"] = None
+
+    def __new__(cls, *args, **kwargs):
+        if args and args[0] == DELETED_SENTINEL and cls._sentinel:
+            return cls._sentinel
+        obj = super().__new__(cls)
+        super().__init__(obj)
+        if args and args[0] == DELETED_SENTINEL:
+            obj._name = "(DELETED)"
```
cmgrote, line 24: Third place the string `(DELETED)` has appeared (-> constant)

pyatlan/model/custom_metadata.py
```diff
@@ -70,7 +94,7 @@ def is_set(self, key: str):
 
     @property
     def business_attributes(self) -> dict[str, Any]:
-        """Returns a dict containing the metadat set with the human readable set name and property names resolved
+        """Returns a dict containing the metadat set with the human-readable set name and property names resolved
```
cmgrote, line 97: `metadat` -> `metadata`

pyatlan/model/custom_metadata.py
```diff
@@ -87,12 +111,16 @@ def __init__(self, business_attributes: Optional[dict[str, Any]]):
...
-            attribs._modified = False
+            try:
+                cm_name = CustomMetadataCache.get_name_for_id(cm_id)
+                attribs = CustomMetadataDict(name=cm_name)
+                for attr_id, properties in cm_attributes.items():
+                    attr_name = CustomMetadataCache.get_attr_name_for_id(cm_id, attr_id)
+                    attribs[attr_name] = properties
+                attribs._modified = False
+            except NotFoundError:
+                cm_name = "(DELETED)"
```
cmgrote, line 122: constant...

tests/integration/test_client.py
```diff
@@ -346,3 +366,56 @@ def test_workflow_find_by_type(client: AtlanClient):
...
+        )
+    )
+
+    assert results.total_count > 0
+    count = len(results.current_page())
+    assert count > 0 and count <= size
+
+
+@pytest.mark.order(after="test_audit_find_by_user")
+def test_audit_find_by_guid(client: AtlanClient, audit_info: AuditInfo):
```
cmgrote, line 408: Would be awesome to see some deeper integration tests that flex parsing back the different kinds of audit history entries that could exist based on preceding tests (see `SQLAssetTest` in Java SDK as an example); rather than just checking there is any audit log entry at all (without introspecting whether its contents match expectations).

ErnestoLoma: I'm not sure that I'm following you. I see in the SQLAssetTest you are doing a searchByParentQN that seems to be looking for assets you created in the previous tests in the file. I'm not creating any assets. I'm just looking at whatever the most recent audit logs happen to be. Do you mean I should create an asset and then look for it by_guid and by_qualified_name?

ErnestoLoma: Never mind. I see you have a searchAuditLogByGuid test. I will look at that.

cmgrote: Or add these integration tests into one of the other existing integration test modules, like `sql_asset_test.py` — that way we don't necessarily have to go through the process of creating even more connections, assets, tear-downs, etc (adding more execution time to the integration tests); instead we're just running these searches at the end of what's already there to confirm we get back what's expected.

pyatlan/model/core.py
```diff
@@ -17,6 +17,8 @@
 
 from pyatlan.model.enums import AnnouncementType, EntityStatus
 
+DELETED_SENTINEL = "DELETED_SENTINEL"
```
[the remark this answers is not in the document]
ErnestoLoma: One end's up being the value of the tag. Same as you are using. One is a unique string I used that was unlikely to be the value of a tag. Originally I was going to make it something more unique like "DeLeTeD_SeNtIn3l" to make it very unlikely to be a tag name. I have done that now.

pyatlan/model/audit.py
```diff
@@ -0,0 +1,272 @@
...
+            _from=_from,
+        )
+        return AuditSearchRequest(dsl=dsl)
+
+    @classmethod
+    def by_qualified_name(
+        cls, type_name: str, qualified_name: str, *, size: int = 10, _from: int = 0
+    ) -> "AuditSearchRequest":
+        """
+        Create an audit search request ffor the last changes to an asset, by its qualifiedName.
```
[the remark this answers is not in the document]
ErnestoLoma: Sorry. I thought I had fixed this.

pyatlan/model/audit.py
```diff
@@ -0,0 +1,272 @@
...
+    created: datetime
+    user: str
+    action: AuditActionType
+    details: Optional[Any]
+    event_key: str
+    entity: Optional[Any]
+    type: Optional[Any]
+    detail: Optional[Union[CustomMetadataAttributesAuditDetail, AtlanTag, Asset]]
+    entity_detail: Optional[Asset]
+    headers: Optional[dict[str, str]]
```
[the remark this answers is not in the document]
ErnestoLoma: As far a I know there's not a way to add doc strings. I have add Field descriptions.

HISTORY.md
```diff
@@ -1,3 +1,13 @@
+## 1.3.0 (October 5, 2023)
```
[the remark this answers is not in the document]
0xquark: A good suggestion! 🚀  I've seen a lot of releases following similar pattern.

0xquark (approved):

PR #143, opened 2023-10-06
Correct type of is_locked attrbuted in custom metadata def

ErnestoLoma:


cmgrote (changes requested):
There's still code like this in the various static helper methods:

```
return CustomMetadataDef.Options(
    emoji=emoji, logo_type="emoji", is_locked="true" if locked else "false"
)
```

Shouldn't this now be just:

```
return CustomMetadataDef.Options(
    emoji=emoji, logo_type="emoji", is_locked=locked
)
```

(?)

ErnestoLoma:
Made requested changes.

cmgrote (approved):

PR #144, opened 2023-10-08
AM-255: Manage API Assets 

0xquark:
PR to add API Assets: APISpec and APIPath (https://developer.atlan.com/patterns/create/api/?h=api+assets)
Please take a look and tell me if there's anything that needs to be fixed or made better.

P.S - A lot of commits due to me branching out this branch from AM-542 Therefore reverted those commits.

CC: @cmgrote @ErnestoLoma


pyatlan/generator/templates/methods/attribute/a_p_i_path.jinja2
```diff
@@ -0,0 +1,27 @@
+
+        @classmethod
+        # @validate_arguments()
+        def create(cls, *, name: str, connection_qualified_name: str, apiPathRawURI: str, apiSpecQualifiedName: str) -> {{ entity_def.name }}.Attributes:
```
cmgrote, line 4: I don't think we need all of these parameters:

- The `connection_qualified_name` can be calculated from the `api_spec_qualified_name`.
- The `name` can be calculated from the `api_path_raw_uri` (this must be unique within a spec anyway).

See the `creator(String, String)` method in the Java SDK's `com.atlan.model.assets.APIPath` for the logic 🙏

0xquark: I understand but according to the documentation isn't `If the name of your path does no give the URI of the endpoint it represents, be sure to also specify the apiPathRawURI.` means that name and api_path_raw_uri can have different value. How do i integrate something like this?

0xquark: Alright! Understood, Thanks! Fixed this in the code 👍

pyatlan/generator/templates/methods/asset/a_p_i_path.jinja2
```diff
@@ -0,0 +1,11 @@
+
+    @classmethod
+    # @validate_arguments()
+    def create(cls, *, apiPathRawURI: str, apiSpecQualifiedName: str) -> {{ entity_def.name }}:
```
cmgrote, line 4: Need to go from `lowerCamelCase` to `snake_case`, though 🙏

pyatlan/generator/templates/methods/attribute/a_p_i_path.jinja2
```diff
@@ -0,0 +1,27 @@
+
+        @classmethod
+        # @validate_arguments()
+        def create(cls, *, apiPathRawURI: str, apiSpecQualifiedName: str) -> {{ entity_def.name }}.Attributes:
```
cmgrote, line 4: Casing for variables again...

cmgrote (changes requested):
Just some casing consistency changes, plz 🙏

pyatlan/generator/templates/methods/asset/a_p_i_path.jinja2
0xquark: Thanks! I'd keep that in mind for future as well 👍

pyatlan/generator/templates/methods/asset/a_p_i_path.jinja2
```diff
@@ -0,0 +1,11 @@
+
+    @classmethod
+    # @validate_arguments()
+    def create(cls, *, path_raw_u_r_i: str, spec_qualified_name: str) -> {{ entity_def.name }}:
```
cmgrote, line 4: Ok, last one 🤞 — for these variable names (since we're writing them rather than relying on some generated code to pick them up like in the filename itself), let's not do the spaced-out naming for `snake_case`: so `path_raw_uri` instead of `path_raw_u_r_i` 🙏

cmgrote (changes requested):
One more final little change on the snake-casing within the file 🙏

pyatlan/generator/templates/methods/asset/a_p_i_path.jinja2
0xquark: I get this error while just changing the characters like this on the `a_p_i_path` unit test
`>       assert sut.api_path_raw_uri == API_PATH_RAW_URI
E       AttributeError: 'APIPath' object has no attribute 'api_path_raw_uri'
`
Not sure why! Since i already defined the attributes in the attribute template
```
return {{ entity_def.name }}.Attributes(
                api_path_raw_uri=path_raw_uri,
   ```

0xquark: Solved it! My bad `api_path_raw_u_r_i` was defined as attribute as part of the template.

cmgrote (approved):
Looks good now!

PR #149, opened 2023-10-12
Am 148 SDK hygiene

ErnestoLoma:
Eliminate use of List and Dict
Create additional client classes:
- GroupClient
- RoleClient
- TokenClient
- TypeDefClient
- UserClient
AssetClient still to come


cmgrote (approved):
Just a minor cleanup suggestion on the history file — otherwise looks good 👍

PR #152, opened 2023-10-17
Translate `classificationNames` as Atlan tag names

ErnestoLoma:


pyatlan/generator/templates/referenceable_methods.jinja2
```diff
@@ -22,3 +22,14 @@
...
     def flush_custom_metadata(self):
         self.business_attributes = self._metadata_proxy.business_attributes
+
+    @property
+    def atlan_tag_names(self) -> list[str]:
+        from pyatlan.cache.atlan_tag_cache import AtlanTagCache
+
+        if self.classification_names:
+            return [
+                AtlanTagCache.get_name_for_id(tag_id) or f"{tag_id}-archived"
```
cmgrote, line 32: Curious why we use `{tag_id}-archived` here as a fallback value. Didn't we use some other sentinel value for deleted elsewhere? (Should be consistent here, I think.)

cmgrote (changes requested):

tests/unit/test_model.py
```diff
@@ -869,3 +874,24 @@ def test_validate_single_required_field_with_bad_values_raises_value_error(
...
+        return ""
+
+    monkeypatch.setattr(
+        pyatlan.cache.atlan_tag_cache.AtlanTagCache,
+        "get_name_for_id",
+        get_name_for_id,
+    )
+
+    referenceable = Referenceable()
+    referenceable.classification_names = [tag_id, "456"]
```
cmgrote, line 895: Is there a positive test here, too (`123` matching a defined tag and thus getting back `Issue` as a name)? 

This only looks like the negative test (`456` not matching any defined tag, and thus getting back the deleted sentinel)

ErnestoLoma: Yes the assert on line 897 says that referenceable.atlan_tag_names == [tag_name, DELETED_]. tag_name is "Issue"

cmgrote (approved):

tests/unit/test_model.py
cmgrote: Ah, now I see it — eyes are blurry from staring at a screen for too long 😵‍💫

PR #153, opened 2023-10-18
Automatically apply placeholder GUIDs to bulk requests

ErnestoLoma:


pyatlan/generator/templates/structs.jinja2
```diff
@@ -11,14 +11,15 @@ from pyatlan.model.enums import (
...
-from pyatlan.utils import validate_required_fields
+from pyatlan.utils import init_guid, validate_required_fields
 
 {% for struct in struct_defs %}
 class {{struct.name}}(AtlanObject):
     """Description"""
     {%- if struct.name == "BadgeCondition" %}
     @classmethod
     # @validate_arguments()
+    @init_guid
```
cmgrote, line 22: I don't think there's a GUID on structs — unsure why this would be needed here?

pyatlan/version.txt
```diff
@@ -1 +1 @@
-1.3.3
+1.3.4
```
cmgrote, line 1: Given the new embedded clients are fairly extensive, I'd be tempted to bump this version to 1.4.0 — thoughts?

cmgrote (commented):
Just 2 things to sense-check

pyatlan/generator/templates/structs.jinja2
ErnestoLoma: Ok. I will remove it, but it does check if the class doesn't have a guid it does nothing.

pyatlan/version.txt
ErnestoLoma: Agreed.

cmgrote (approved):

PR #154, opened 2023-10-19
Dx 11 - Add asset batching mechanism

ErnestoLoma:


pyatlan/client/asset.py
```diff
@@ -1421,3 +1422,152 @@ def _get_next_page(self):
...
+            except AtlanError as er:
+                if self._capture_failures:
+                    self._failures.append(
+                        FailedBatch(failed_assets=self._batch, failure_reason=er)
+                    )
+                else:
+                    raise er
+            self._batch = []
+        if response:
+            self._track_repsonse(response=response)
```
cmgrote, line 1558: Minor given its internal, but probably should fix the typo here: `_track_repsonse` -> `_track_response`

cmgrote (changes requested):

pyatlan/client/asset.py
ErnestoLoma: done

cmgrote (approved):

PR #155, opened 2023-10-24
DX-14 : Manage GCS Assets

0xquark:


tests/integration/gcs_asset_test.py
```diff
@@ -0,0 +1,258 @@
...
+            # If it is not found, it was successfully deleted (purged), so we
+            # do not need to look for it any further
+            print("Asset no longer exists.")
+        except AtlanError:
+            leftovers.append(one)
+    if leftovers:
+        if retry_count == 20:
+            raise ErrorCode.RETRY_OVERRUN.exception_with_parameters()
+        time.sleep(2)
+        _retrieve_and_check(client, leftovers, retry_count + 1)
```
ErnestoLoma, line 57: This code is repeated in multiple integration tests. Please refactor it into tests/integretaion/utils.py.

tests/integration/gcs_asset_test.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+@pytest.fixture(scope="module")
+def connection(client: AtlanClient) -> Generator[Connection, None, None]:
+    result = create_connection(
+        client=client, name=MODULE_NAME, connector_type=CONNECTOR_TYPE
+    )
+    yield result
+    delete_asset(client, guid=result.guid, asset_type=Connection)
```
ErnestoLoma, line 66: This fixture is defined in other integration tests. No need to define it agian unless you need to do something different than in the other tests. Please remove this block and check if other tests have duplicated it. It should only be defined in one place. Please move it to tests/integration/connection.py and add connection.py to conftest.py

tests/unit/model/g_c_s_bucket_test.py
ErnestoLoma, line 1: Please change the name to gcs_bucket_test.py

tests/unit/model/g_c_s_object_test.py
ErnestoLoma, line 1: Please change the name to gcs_object_test.py

ErnestoLoma (changes requested):

tests/integration/gcs_asset_test.py
0xquark: As this fixture requires `MODULE_NAME` and `CONNECTOR_TYPE` from particular assets, How do i pass this to a fixture? afaik we can't call the fixtures directly and send arguments.

ErnestoLoma (approved):

PR #160, opened 2023-10-31
DX-257: Publish pre-built container image with embedded Python SDK

0xquark:


.github/workflows/python-publish.yaml
```diff
@@ -48,3 +48,22 @@ jobs:
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          push: true
+          tags: ghcr.io/${{ github.repository }}:${{ needs.deploy.outputs.version }}
```
cmgrote, line 67: I don't think this line is going to work without having set up the `version` output in an earlier step.

See these lines for an example of how this output is generated and set on a variable:
- https://github.com/atlanhq/atlan-java/blob/fc2f7ed0af250de4b72c51ac1572e1e8dede5902/.github/workflows/release.yml#L14-L15
- https://github.com/atlanhq/atlan-java/blob/fc2f7ed0af250de4b72c51ac1572e1e8dede5902/.github/workflows/release.yml#L22-L25

Not sure how we'd get the version from Python (the second bullet) (?)

Dockerfile
cmgrote, line 1: We should create a top-level directory in the repo called something like `container`, inside of which is _only_ this `Dockerfile`. A step in the GitHub Action should be to copy the `requirements.txt` file into the same directory during the run.

^^ This ensures that when building the container image, _only_ the `requirements.txt` and any installed Python libraries are included in the container image. Otherwise, _everything_ that's in the repo will be copied into a layer in the container image (increasing the image's size).

cmgrote (changes requested):

.github/workflows/python-publish.yaml
```diff
@@ -64,6 +64,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           push: true
-          tags: ghcr.io/${{ github.repository }}:${{ needs.deploy.outputs.version }}
-          context: .
-          file: ./Dockerfile
+          tags: ghcr.io/atlanhq/atlan-python:$(cd .. && cat pyatlan/version.txt)
```
0xquark, line 67: Not sure in which directory this runs in! I am assuming it to be inside containers directory.
Would need to run the workflow to check this

cmgrote (approved):

containers/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.9.16-slim-buster
+
+ARG VERSION
+
+WORKDIR /app
+
+RUN pip3 install requests>=2.24 pydantic==1.10.12 jinja2==3.1.2 networkx==3.1 tenacity==8.2.3
```
ErnestoLoma, line 7: I don't believe that this is needed. I think whatever pyatlan needs will be installed when you install it.

ErnestoLoma (changes requested):

containers/Dockerfile
0xquark: Aight 👍

ErnestoLoma (approved):

PR #166, opened 2023-11-03
Workflow: Add variable to use as a tag for image

0xquark:


cmgrote (approved):
Hopefully this approach works 🤞

PR #170, opened 2023-11-06
Dx 236 Add glossary hierarchy traversal logic

ErnestoLoma:


pyatlan/client/asset.py
```diff
@@ -1256,6 +1258,40 @@ def find_term_by_name(
...
+            .include_on_results(AtlasGlossaryCategory.PARENT_CATEGORY)
+            .page_size(20)
+            .sort(AtlasGlossaryCategory.NAME.order(SortOrder.ASCENDING))
+        )
+        for field in attributes:
+            search.include_on_results(field)
+        request = search.to_request()
+        response = self.search(request)
+        for category in filter(
+            lambda a: isinstance(a, AtlasGlossaryCategory), response
```
cmgrote, line 1282: Better that we explicitly search for only `AtlasGlossaryCategory` type in the FluentSearch, to push-down the filtering into the search itself

cmgrote (changes requested):
`GlossaryCategory.select()` in Java SDK pushes down a search to _only_ retrieve `AtlasGlossaryCategory` objects — I think here we're retrieving all kinds of assets and then only filtering through the lambda when iterating (?)

pyatlan/client/asset.py
```diff
@@ -1256,6 +1258,40 @@ def find_term_by_name(
...
+    ) -> CategoryHierarchy:
+        from pyatlan.model.fluent_search import FluentSearch
+
+        if attributes is None:
+            attributes = []
+        top_categories: set[str] = set()
+        category_dict: dict[str, AtlasGlossaryCategory] = {}
+        search = (
+            FluentSearch.select()
+            .where(AtlasGlossaryCategory.ANCHOR.eq(glossary.qualified_name or ""))
```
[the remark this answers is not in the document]
ErnestoLoma: @cmgrote Are you saying I should raise an exception then?

pyatlan/client/asset.py
ErnestoLoma: @cmgrote Ok but in the JDK you have the following:
` GlossaryCategory.select(client)
                .where(GlossaryCategory.ANCHOR.eq(getQualifiedName()))
                .includeOnResults(GlossaryCategory.PARENT_CATEGORY)
                .includesOnResults(attributes == null ? Collections.emptyList() : attributes)
                .pageSize(20)
                .sort(GlossaryCategory.NAME.order(SortOrder.Asc))
                .stream()
                .filter(a -> a instanceof GlossaryCategory)`
Am I missing something?  I thought this meant the filtering was occurring on the stream and not the search so that the search would retrieve all types of assets and they would be filtered in the stream. So doesn't an equivalent change need to be done there?

ErnestoLoma: done

pyatlan/client/asset.py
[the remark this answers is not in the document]
ErnestoLoma: done

pyatlan/client/asset.py
cmgrote: Yeah, I should probably drop that line from the Java SDK. The hope with the filter there was more that it would make the explicit cast to a term that comes in the `forEach` redundant, but naturally it's still required in Java (not so in Kotlin).

cmgrote (approved):

PR #172, opened 2023-11-08
DX-276 Custom metadata changes

ErnestoLoma:


0xquark (approved):

pyatlan/model/typedef.py
```diff
@@ -233,11 +239,19 @@ def get_valid_values(self) -> Optional[list[str]]:
...
             description="Optional description of the attribute."
         )
         applicable_entity_types: Optional[str] = Field(
-            description="Set of entities on which this attribute can be applied.",
+            description="Set of entities on which this attribute can be applied. "
+            "Note: generally this should be left as-is. Any overrides should instead be applied through "
+            "one or more of applicable_asset_types}, applicable_glossary_types}, or "
+            "applicable_other_asset_types}.",
         )
         custom_applicable_entity_types: Optional[str] = Field(
```
cmgrote, line 256: Let's at least add to the description of this one that it is deprecated, to give some notice to folks that they shouldn't use this field anymore.

pyatlan/model/typedef.py
```diff
@@ -390,6 +432,158 @@ def create(
...
+        Set of entities on which this attribute can be applied.
+        Note: generally this should be left as-is. Any overrides should instead be applied through
+        one or more of applicable_asset_types, applicable_glossary_types, or applicable_other_asset_types.
+        """
+        if self.options and self.options.applicable_entity_types:
+            return set(json.loads(self.options.applicable_entity_types))
+        return set()
+
+    @applicable_entity_types.setter
+    def applicable_entity_types(self, entity_types: set[str]):
```
cmgrote, line 461: Nice — so users can still send in a set, even though it's marshalled into JSON as a string 👍

cmgrote (changes requested):
One very minor change — otherwise looks good!

pyatlan/model/typedef.py
ErnestoLoma: Done

pyatlan/model/typedef.py
ErnestoLoma: Done

cmgrote (approved):

PR #175, opened 2023-11-13
DX-297 In-cluster optimisation to use services URLs directly

ErnestoLoma:


pyatlan/client/atlan.py
```diff
@@ -262,26 +262,45 @@ def _call_api_internal(self, api, path, params, binary_data=None):
...
+            LOGGER.debug("------------------------------------------------------")
+            LOGGER.debug("Call         : %s %s", api.method, path)
+            LOGGER.debug("Content-type_ : %s", api.consumes)
+            LOGGER.debug("Accept       : %s", api.produces)
+            LOGGER.debug("Request ID   : %s", request_id)
         return self._call_api_internal(api, path, params)
 
+    def _create_path(self, api: API):
+        if self.base_url == "INTERNAL":
+            return os.path.join(api.endpoint.service, api.path)
```
cmgrote, line 279: Does this use an OS-specific path separator when joining the components together? (If so, won't this use `\` on Windows, and thus not be a valid URL?)

cmgrote (commented):
Just want to confirm the path construction for the endpoint is platform-agnostic

pyatlan/client/atlan.py
ErnestoLoma: Good catch. The use of os.path.join was inherited from the apache atlan API.  I have changed it to use urljoin.

cmgrote (approved):

PR #177, opened 2023-11-14
Changed AuthPolicy.create_for_modification to raise NotImplmentedError

ErnestoLoma:


ErnestoLoma:
Used wrong branch as basis

PR #178, opened 2023-11-14
Changed AuthPolicy.create_for_modification to raise NotImplmentedError

ErnestoLoma:


0xquark (commented):
I don't think the files under `docs/` are meant to be committed

ErnestoLoma:
> I don't think the files under `docs/` are meant to be committed

There are already files in docs. @cmgrote ?

pyatlan/model/enums.py
```diff
@@ -2060,3 +2060,12 @@ class SourceCostUnitType(str, Enum):
     CREDITS = "Credits"
     BYTES = "bytes"
     SLOT_MS = "slot-ms"
+
+
+class jsdk_DataQualityType_RVUuI(str, Enum):
```
cmgrote, line 2065: This is a generated enum from integration tests — should revert to the last version of `enums.py` without this class.

cmgrote (changes requested):
`docs` should stay — that's to generate a separate JavaDocs-equivalent site that just gives info about the entire Python SDK (its classes, operations, etc).

pyatlan/model/enums.py
ErnestoLoma: done

cmgrote (approved):

PR #179, opened 2023-11-15
DVX-63 Update AuditSearch.by_qualified_name

ErnestoLoma:


cmgrote (approved):
I'm not really clear on how this fixed things (what was the issue in the end?) but looks like it works fine with the tests (?)

ErnestoLoma:
> I'm not really clear on how this fixed things (what was the issue in the end?) but looks like it works fine with the tests (?)

@cmgrote The issue that I thought existed originally didn't exist. I'm not sure what I was looking at but during testing I did find an issue where it was sometimes classifying an asset as a deleted tag. That is what I fixed by making the tag definition a bit stricter.

PR #181, opened 2023-11-15
DVX-40: Ability to set icon during glossary creation

0xquark:


pyatlan/generator/templates/methods/asset/atlas_glossary.jinja2
```diff
@@ -8,6 +8,7 @@
     @classmethod
     # @validate_arguments()
     @init_guid
-    def create(cls, *, name: StrictStr)->{{ entity_def.name }}:
+    def create(cls, *, name: StrictStr, icon: Optional[AtlanIcon] = None)->{{ entity_def.name }}:
         validate_required_fields(["name"], [name])
-        return AtlasGlossary(attributes=AtlasGlossary.Attributes.create(name=name))
+        icon_str = icon.value if icon is not None else None
```
cmgrote, line 13: Looks like we're doing this translation from enum to string value in both the `AtlasGlossary::create()` and the `AtlasGlossary.Attributes::create()` methods (?)

We should only do the translation in one place (probably `AtlasGlossary.Attributes`) — so both `create()` methods take an enum value, the `AtlasGlossary::create()` just passes that enum value forward to the `AtlasGlossary.Attribute::create()`, and only this latter one does the translation to the string value.

cmgrote (changes requested):

cmgrote (changes requested):
Revision looks good, just need to get rid of an extraneous generated enum

cmgrote (approved):

PR #182, opened 2023-11-16
DVX-50: Bundle `logging.conf` file into base SDK image

0xquark:


containers/Dockerfile
```diff
@@ -6,4 +6,6 @@ WORKDIR /app
 
 RUN pip3 install pyatlan==${VERSION}
 
+COPY ../pyatlan/logging.conf .
```
cmgrote, line 9: This might have changed in Docker, but at least in the "old days" it wasn't possible to copy files that are outside the path the `Dockerfile` itself is in.

So you'll probably instead need to use the `.github/workflows/...yml` file that builds the container to instead copy the file from the (checked out) repo into the same (temporary) "assembly" directory for the image (same directory the `Dockerfile` is in in the checked-out repo), and then here in the `Dockerfile` you'd just do a `COPY logging.conf .`

Slightly different scenario, but see for example: https://github.com/atlanhq/atlan-java/blob/c0030d1a1f5ca42f5918be6b81798d8e7a47431a/.github/workflows/merge.yml#L140-L146 <— this creates a `container/assembly` directory and downloads a jar file built earlier in the build process into it, that can then be copied within the `Dockerfile` itself.

cmgrote (changes requested):

cmgrote (approved):

PR #183, opened 2023-11-19
Add `gdsid` as optional parameter

0xquark:


pyatlan/model/assets/asset00.py
```diff
@@ -201,7 +201,6 @@ def validate_required(self):
     type_name: str = Field(
         "Referenceable",
         description="Name of the type definition that defines this instance.\n",
-        regex=r"^([A-Z][a-z])+$",
```
cmgrote, line 204: @ErnestoLoma looks like regex piece you'd added is being dropped in this PR... was the regex piece added as part of the code generation? (Feels like it was lost as part of regenerating?)

cmgrote (commented):
GDS ID piece looks good, just need Ernest's feedback on whether this `regex` parameter is no longer needed or if it was a miss on it being part of the code generator (?)

pyatlan/model/assets/asset00.py
ErnestoLoma: @cmgrote I added but then removed it. I guess I forgot to regnerate. It's not needed.

ErnestoLoma (approved):

PR #186, opened 2023-11-20
DVX-65: Adds a tag update method

Aryamanz29:


cmgrote (approved):
Looks right to me — just to confirm, you've run some tests locally?

Aryamanz29:
> Looks right to me — just to confirm, you've run some tests locally?

Yes @cmgrote, I have run both unit and integration tests on these changes ✅

PR #191, opened 2023-11-24
DVX-74: Add data mesh functionality

Aryamanz29:


pyatlan/generator/templates/methods/asset/data_domain.jinja2
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    # @validate_arguments()
+    @init_guid
+    def create(
+        cls,
+        *,
+        name: StrictStr,
+        icon: Optional[AtlanIcon] = None,
+        parent_domain: Optional[DataDomain] = None,
+        parent_domain_guid: Optional[StrictStr] = None,
```
cmgrote, line 11: As discussed, let's change this to qualifiedName rather than GUID

pyatlan/generator/templates/methods/asset/data_domain.jinja2
```diff
@@ -0,0 +1,55 @@
...
+        validate_required_fields(["name", "qualified_name"], [name, qualified_name])
+        # Split the data domain qualified_name to extract data mesh info
+        fields = qualified_name.split("/")
+        if len(fields) != 3:
+            raise ValueError(f"Invalid data domain qualified_name: {qualified_name}")
+        mesh_slug, mesh_abbreviation = fields[-1], fields[-1]
+        # If "guid" of the parent domain is specified
+        if parent_domain_guid:
+            parent_domain = DataDomain()
+            parent_domain.guid = parent_domain_guid
```
cmgrote, line 46: Just be wary here, when we switch to qualified name, it isn't `parent_domain.qualified_name` that needs to be set, but `parent_domain.unique_attributes.qualified_name` (the latter is for a relationship)

pyatlan/generator/templates/methods/asset/data_product.jinja2
```diff
@@ -0,0 +1,59 @@
...
+    # @validate_arguments()
+    @init_guid
+    def create(
+        cls,
+        *,
+        name: StrictStr,
+        assets_dsl: StrictStr,
+        icon: Optional[AtlanIcon] = None,
+        domain: Optional[DataDomain] = None,
+        domain_guid: Optional[StrictStr] = None,
```
cmgrote, line 12: As discussed, let's change this to qualifiedName rather than GUID

pyatlan/generator/templates/methods/asset/data_product.jinja2
```diff
@@ -0,0 +1,59 @@
...
+            name=self.name or "",
+            domain_guid=self.data_domain.guid,
+        )
+
+    @classmethod
+    def create_for_modification(
+        cls: type[SelfAsset],
+        qualified_name: str = "",
+        name: str = "",
+        domain_guid: str = "",
```
cmgrote, line 38: As discussed, let's change this to qualifiedName rather than GUID

pyatlan/generator/templates/methods/asset/data_domain.jinja2
```diff
@@ -0,0 +1,55 @@
...
+            parent_domain_guid=self.parent_domain and self.parent_domain.guid or "",
+        )
+
+    @classmethod
+    def create_for_modification(
+        cls: type[SelfAsset],
+        qualified_name: str = "",
+        name: str = "",
+        parent_domain: Optional[DataDomain] = None,
+        parent_domain_guid: Optional[StrictStr] = None,
```
cmgrote, line 35: As discussed, let's change this to qualifiedName rather than GUID

pyatlan/generator/templates/methods/asset/data_product.jinja2
```diff
@@ -0,0 +1,59 @@
...
+            ["name", "qualified_name", "domain_guid"],
+            [name, qualified_name, domain_guid],
+        )
+        # Split the data domain qualified_name to extract data mesh info
+        fields = qualified_name.split("/")
+        if len(fields) != 3:
+            raise ValueError(f"Invalid data product qualified_name: {qualified_name}")
+        mesh_slug, mesh_abbreviation = fields[-1], fields[-1]
+        domain = DataDomain()
+        domain.guid = domain_guid
```
cmgrote, line 50: As mentioned above, the move to qualified name would make this `domain.unique_attributes.qualified_name` since it will be used to define a relationship, not `domain.qualified_name`.

pyatlan/model/search.py
```diff
@@ -1848,6 +1849,11 @@ class IndexSearchRequest(SearchRequest):
         alias="allowDeletedRelations",
     )
 
+    def get_dsl_str(self) -> str:
+        dsl_json_str = self.json(by_alias=True, exclude_none=True)
+        dsl_query_dict = dict(query=loads(dsl_json_str))
+        return dumps(dsl_query_dict)
```
cmgrote, line 1855: As discussed, let's embed calling this into the `create()` method (i.e. have the `create()` method take a `FluentSearch` object or `IndexSearchRequest` object itself, and translate it to a string inside the `create()` method), rather than relying on folks to remember to call it themselves.

cmgrote (changes requested):

pyatlan/model/search.py
Aryamanz29: I tried using `FluentSearch`, but it didn't work due to a circular import error. Therefore, I opted to use the `IndexSearchRequest` object directly and translated it into string inside the `create()` method of the `DataProduct`.

pyatlan/generator/templates/methods/asset/data_domain.jinja2
```diff
@@ -0,0 +1,42 @@
...
+        cls: type[SelfAsset],
+        qualified_name: str = "",
+        name: str = "",
+    ) -> SelfAsset:
+        validate_required_fields(["name", "qualified_name"], [name, qualified_name])
+        # Split the data domain qualified_name to extract data mesh info
+        fields = qualified_name.split("/")
+        # for domain and subdomain
+        if len(fields) not in (3, 5):
+            raise ValueError(f"Invalid data domain qualified_name: {qualified_name}")
```
cmgrote, line 33: I don't think this check is needed, as:

1. I think we allow any number of nested domains (not just 2 levels) — this would limit to only 2 levels.
2. The slug and abbreviation can always be taken from the last token, irrespective of how deeply nested the domain is (or even if it's top-level).

pyatlan/generator/templates/methods/asset/data_product.jinja2
```diff
@@ -0,0 +1,47 @@
...
+        name: str = "",
+    ) -> SelfAsset:
+        validate_required_fields(
+            ["name", "qualified_name"],
+            [name, qualified_name],
+        )
+        # Split the data product qualified_name to extract data mesh info
+        fields = qualified_name.split("/")
+        if len(fields) != 3:
+            raise ValueError(f"Invalid data product qualified_name: {qualified_name}")
```
cmgrote, line 38: We could leave this, but I'm a bit wary if for some reason our qualifiedName structure changes after this initial closed preview then we'll need to change this as well — I'd probably opt for dropping this check, too.

pyatlan/model/assets/asset00.py
```diff
@@ -6372,6 +6424,636 @@ def create(cls, *, asset: Asset, name: str, link: str) -> Link.Attributes:
...
+                qualified_name = f"{parent_domain_qualified_name}/domain/{mesh_name}"
+            icon_str = icon.value if icon is not None else None
+            return DataDomain.Attributes(
+                name=name,
+                parent_domain=parent_domain,
+                mesh_slug=mesh_name,
+                mesh_abbreviation=mesh_name,
+                qualified_name=qualified_name,
+                icon=icon_str,
+            )
```
cmgrote, line 6742: I was expecting to see this attributes-level stuff generated, so this is good — but I don't see the attributes-level template that defines this logic for the code generation? (Are you sure you've committed everything as part of this PR?)

pyatlan/model/assets/asset00.py
```diff
@@ -6372,6 +6424,636 @@ def create(cls, *, asset: Asset, name: str, link: str) -> Link.Attributes:
...
+            camel_case_name = to_camel_case(name)
+            return DataProduct.Attributes(
+                name=name,
+                data_product_assets_d_s_l=assets_dsl,
+                data_domain=domain,
+                mesh_slug=camel_case_name,
+                mesh_abbreviation=camel_case_name,
+                qualified_name=f"default/product/{camel_case_name}",
+                icon=icon_str,
+            )
```
cmgrote, line 7048: As above, I don't see the template that's used to generate this logic in the PR (?)

cmgrote (changes requested):
Few more little changes... Can you also merge the latest `main` to your `DVX-74` branch to ensure it's up-to-date with any other changes that have since been merged? 🙏

pyatlan/generator/templates/methods/asset/data_domain.jinja2
Aryamanz29: Right, I'll remove that, thanks!

PR #195, opened 2023-12-04
Added check for archived attributes

ErnestoLoma:


cmgrote (changes requested):
I'm not sure just checking for `archived` somewhere in the name of the custom attribute will be robust enough — prevents any actual (non-archived) custom attribute from having that word anywhere in its name.

Wondering if instead we should do a check during the `__init__` e.g. at line 39?
- From the `CustomMetadataCache` see if the attribute is actually archived (there's a property on every `AttributeDef` of `isArchived`) — this'll be more robust than depending on a naming convention for the attribute's name
- Only add it to the set of names for the custom metadata dict if it is not actually archived

PR #196, opened 2023-12-05
DVX-106 Raise exception if Asset.create_modification is used directlyInitial implementation

ErnestoLoma:


pyatlan/errors.py
```diff
@@ -463,6 +463,13 @@ class ErrorCode(Enum):
         "Please use purge if you wish to remove assets of this type",
         InvalidRequestError,
     )
+    METHOD_CAN_NOT_BE_INVOKED_ON_ASSET = (
+        400,
+        "ATLAN-PYTHON-400-053",
+        "This method can not be invoked on the Asset class. Please invoke on a specific asset type",
```
cmgrote, line 469: `can not` -> `cannot`

cmgrote (commented):
Very minor, but suggested change is to use what I believe to be far more common grammar.

cmgrote (approved):

PR #199, opened 2023-12-14
Modify constants like Referenceable.TYPE_NAME to use InternalKeywordField, InternalKeywordTextField or InternalNumericField

ErnestoLoma:


cmgrote (commented):
Looks good, but what about the `in_lineage` translation piece to first check whether the protocol is implemented and thus take the internal field name, and only fallback to another field name if that internal field name is not present (protocol not implemented)?

ErnestoLoma:
@cmgrote For all classes the internal_field_name property is being used. InternalKeywordTextField, InternalKeywordfield and InternalNumericField override that to return the value self._internal_field_name while other classes return the value of self.atlan_field_name.

cmgrote (approved):

PR #201, opened 2023-12-18
Adds method to create unique-per-asset links based on title

cmgrote:


pyatlan/generator/templates/methods/asset/asset.jinja2
```diff
@@ -7,6 +7,21 @@
...
+        if bool(
+            self.unique_attributes
+            and self.unique_attributes.get("qualified_name")
+            and str(self.unique_attributes.get("qualified_name")).strip()
+        ):
+            return self.ref_by_qualified_name(
+                str(cast(dict[str, Any], self.unique_attributes).get("qualified_name"))
+            )
+        raise ErrorCode.MISSING_REQUIRED_RELATIONSHIP_PARAM.exception_with_parameters(self.type_name, "guid, qualifiedName")
+
```
ErnestoLoma, line 24: The bools are superfluous and I believe you are using str to get around the mypy warning that the qualified_name is  Optional. I think the following is cleaner:
```
        if self.guid and self.guid.strip():
            return self.ref_by_guid(self.guid)
        if self.qualified_name and self.qualified_name.strip():
            return self.ref_by_qualified_name(self.qualified_name)
        if (
            self.unique_attributes
            and (qualified_name := self.unique_attributes.get("qualified_name"))
            and qualified_name.strip()
        ):
            return self.ref_by_qualified_name(qualified_name)
        raise ErrorCode.MISSING_REQUIRED_RELATIONSHIP_PARAM.exception_with_parameters(
            self.type_name, "guid, qualifiedName"
        )
```

ErnestoLoma (changes requested):
One change

pyatlan/generator/templates/methods/asset/asset.jinja2
cmgrote: Yes, thanks! Seemed a bit ridiculous before 😬

ErnestoLoma (approved):
Looks good

PR #203, opened 2023-12-20
DVX-118: Add support for workflow packages

Aryamanz29:
## Todos:

**Need to add support for following packages:**

- [x] Snowflake crawler
- [x] DBT
- [x] Kafka
- [x] Power BI
- [x] Glue
- [x] Tableau

**Other tasks:**

- [x] Add tests.
- [ ] Update developer portal docs (WIP).

### Snippets:

```py

from pyatlan.model.packages.snowflake_crawler import SnowflakeCrawler
client = AtlanClient()

crawler = (
    SnowflakeCrawler(
        connection_name="<connection-name>",
        admin_roles=[RoleCache.get_id_for_name("$admin")],
        admin_groups=None,
        admin_users=None,
    )
    .basic_auth(
        username="<username>",
        password="<password>",
        role="<role>",
        warehouse="<warehouse>",
    )
    .information_schema(hostname="<hostname>")
    .include(assets={<assets>})
    .lineage(True)
    .tags(False)
    .to_workflow()
)

print(client.workflow.run(crawler))
```

### DIFF
--- a/pyatlan/model/packages/base/crawler.py
+++ b/pyatlan/model/packages/base/crawler.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Optional
 
 from pyatlan import utils
 from pyatlan.cache.role_cache import RoleCache
+from pyatlan.errors import ErrorCode
 from pyatlan.model.assets import Connection
 from pyatlan.model.packages.base.package import AbstractPackage
 
@@ -82,13 +83,18 @@ class AbstractCrawler(AbstractPackage):
 
         :param raw_filter: map keyed by database name with each value being a list of schemas
         :returns: an exact-match filter map string, usable in crawlers include / exclude filters
+        :raises InvalidRequestException: In the unlikely event the provided filter cannot be translated
         """
         to_include: Dict[str, Any] = {}
-        if raw_filter:
+        if not raw_filter:
+            return ""
+        try:
             for db_name, schemas in raw_filter.items():
                 exact_schemas = [f"^{schema}$" for schema in schemas]
                 to_include[f"^{db_name}$"] = exact_schemas
-        return dumps(to_include)
+            return dumps(to_include)
+        except (AttributeError, TypeError):
+            raise ErrorCode.UNABLE_TO_TRANSLATE_FILTERS.exception_with_parameters()
 
     @staticmethod
     def build_flat_filter(raw_filter: Optional[list]) -> str:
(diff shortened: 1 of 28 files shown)


pyatlan/model/packages/base/crawler.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _get_connection(self) -> Connection:
+        """
+        Builds a connection using the provided parameters,
+        which will be the target for the package to crawl assets.
+        """
+        if not self._admin_roles:
+            role = RoleCache.get_id_for_name("$admin")
+            if role:
+                self._admin_roles = [role]
```
cmgrote, line 62: This is tempting to use as a fallback, but possibly changes the intended behavior. It's perfectly permissible that someone will create a connection that only has connection admins as a particular set of users or a group — and will not allow any roles as connection admins.

In this case, this would set everyone with the Admin role (either currently or at some point in the future) as a connection admin, which would violate the desired behavior above.

If a user wants to set all Admins as connection admins, they therefore need to be explicit rather than us doing it automatically every time the provided list of admin roles is empty.

tests/unit/data/package_requests/confluent_kafka_direct.json
```diff
@@ -0,0 +1,114 @@
...
+        }
+      }
+    },
+    "payload": [
+      {
+        "parameter": "credentialGuid",
+        "type": "credential",
+        "body": {
+          "name": "default-confluent-kafka-123456-0",
+          "host": "test-bootstrap-server",
```
cmgrote, line 102: Are you sure the host doesn't include the `:9092` at the end? Thought I had seen that this is in both the end of the host and in the separate port field...

cmgrote (changes requested):
Looks great overall — just a couple things to check

tests/unit/data/package_requests/confluent_kafka_direct.json
Aryamanz29: Thanks for catching that! I missed it in the tests 👍

pyatlan/model/packages/base/crawler.py
Aryamanz29: Understood. I have removed that condition. Thank you! 👍

Aryamanz29:
I have rebased this PR with [the changes from PR #206](https://github.com/atlanhq/atlan-python/pull/206) to implement the fix [`cd4c764` (#203)](https://github.com/atlanhq/atlan-python/pull/203/commits/cd4c7647381cc20b68b79ea47a0420c845ade7a1)

cmgrote (approved):

PR #207, opened 2023-12-29
Fixes workflow client `monitor` method

Aryamanz29:
The `WorkflowResponse` **metadata.name** provides the name of the workflow. Therefore, when calling 
`/api/service/runs/indexsearch`, we should find the latest workflow run by passing its name it's name to `spec.workflowTemplateRef.name.keyword` to obtain the desired results.


cmgrote (approved):
Yup, looks like what we're doing in the Java SDK 👍

PR #212, opened 2024-01-05
DVX-131: Add ability to update certificate, announcement for GlossaryTerm and GlossaryCategory

Aryamanz29:
**Note:** Initially, I attempted to implement this functionality using third-party packages, such as [multimethod](https://github.com/coady/multimethod), which offer method overloading. [Despite Python not natively supporting](https://stackoverflow.com/a/10202988) method overloading like some other languages, I encountered issues with type hints and significant code duplication. Therefore, I opted for achieving a similar outcome using default values and the `*args` and `**kwargs` syntax.

### DIFF
--- a/pyatlan/client/asset.py
+++ b/pyatlan/client/asset.py
@@ -7,7 +7,7 @@ import logging
 import time
 from abc import ABC
 from enum import Enum
-from typing import Generator, Iterable, Optional, Type, TypeVar, Union
+from typing import Generator, Iterable, Optional, Type, TypeVar, Union, overload
 from warnings import warn
 
 import requests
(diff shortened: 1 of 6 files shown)


pyatlan/client/asset.py
```diff
@@ -850,40 +880,33 @@ def update_certificate(
...
-        )
-        response = AssetMutationResponse(**raw_json)
-        if assets := response.assets_partially_updated(asset_type=asset_type):
-            return assets[0]
-        if assets := response.assets_updated(asset_type=asset_type):
-            return assets[0]
-        return None
-
     @validate_arguments()
     def remove_certificate(
```
ErnestoLoma, line 888: Please use the @overload annotation. See [PEP 3124](https://peps.python.org/pep-3124/) for more info. You will need to create another method with the original signature. workflow.py also uses this annotation.

ErnestoLoma (changes requested):

pyatlan/client/asset.py
Aryamanz29: Hi @ErnestoLoma,

Thanks for the review. I've added the `@overload` decorator, following the example in `workflow.py`. Just a note, similar to `workflow.py`, I've kept type checks [within the method implementation](https://github.com/atlanhq/atlan-python/blob/8b1a2d0cc933e62a63ba009322255059f075c7ea/pyatlan/client/workflow.py#L121-L138). Could you please confirm that my new changes look good?

cmgrote (commented):
I defer to @ErnestoLoma 's knowledge here and final approval 🙏

ErnestoLoma (approved):
Looks good

PR #214, opened 2024-01-09
DVX-155: Always include a sort by GUID as the final criteria

Aryamanz29:


cmgrote:
> It seems like sorting by `GUID` doesn't work with audit search and search log 🤔. The [integration tests fail](https://github.com/atlanhq/atlan-python/actions/runs/7459569913/job/20295859699?pr=214) with a **400** bad request. I think we should exclude sorting by `GUID` for those search requests.

There won't necessarily be unique record in these other indexes (audit search for example will have many records per asset GUID, one record for each change to the asset) — but sorting by it still shouldn't cause a failure. The issue is more likely because in these different indexes the GUID field has different names. In indexsearch it is `__guid`, but in audit search it is (iirc) `entityId`, and in search log it is probably something else still (don't recall what off the top of my head).

pyatlan/model/search.py
```diff
@@ -1829,6 +1827,25 @@ def validate_query(cls, v, values):
...
+        for option in sort:
+            if option.field and option.field in (
+                sort_by_guid,
+                auditsearch_sort_by_guid,
+                searchlog_sort_by_guid,
+            ):
+                missing_guid_sort = False
+                break
+        if missing_guid_sort:
+            sort.append(SortItem(sort_by_guid))
```
cmgrote, line 1846: This still feels a bit strange — we're checking whether the sort includes a GUID-equivalent for indexsearch, auditsearch, and searchlog (all different), but then if that's missing we're only appending the sort by GUID (indexsearch) — the `sort_by_guid` variable here looks like it is hard-coded to the `TermAttributes.GUID.value` (?)

cmgrote (commented):
Not sure if I'm seeing the full context for the highlighted code, but pretty sure it's still only going to work for indexsearch (?)

pyatlan/model/search.py
```diff
@@ -1803,9 +1804,8 @@ class DSL(AtlanObject):
     track_total_hits: bool = Field(True, alias="track_total_hits")
     post_filter: Optional[Query] = Field(alias="post_filter")
     query: Optional[Query]
-    sort: list[SortItem] = Field(
-        alias="sort", default=[SortItem(TermAttributes.GUID.value)]
-    )
+    sort: list[SortItem] = Field(alias="sort", default_factory=list)
+    _req_class_name: ClassVar[str]
```
ErnestoLoma, line 1808: Why is this a class variable? This means there is only one value for the DSL class. This will end up having the last value set. I think this should be a normal required attribute

ErnestoLoma (changes requested):

cmgrote (approved):
LGTM 👍

PR #218, opened 2024-01-15
DVX-68 Adds type checking to AtlanClient save and other methods

Aryamanz29:
**Current limitation:** [pydantic/pydantic/issues/2901](https://github.com/pydantic/pydantic/issues/2901)

I think we have encountered this in the past: [atlanhq/atlan-python/pull/88#discussion_r1260892704](https://github.com/atlanhq/atlan-python/pull/88#discussion_r1260892704)


ErnestoLoma (approved):
Looks good. Thanks for fixing the mypy violations

PR #219, opened 2024-01-15
Add code to support custom package toolkit

ErnestoLoma:
DVX-61 Custom package config toolkit
DVX-62 Custom package runtime toolkit


pyatlan/pkg/models.py
```diff
@@ -0,0 +1,300 @@
...
+            loader=PackageLoader("pyatlan.pkg", "templates")
+        )
+        self._env.globals.update({"isinstanceof": isinstance})
+
+    def create_package(self):
+        self._root_dir.mkdir(parents=True, exist_ok=True)
+        with (self._root_dir / "index.js").open("w") as index:
+            index.write(CustomPackage.indexJS())
+        with (self._root_dir / "index.js").open("w") as index:
+            index.write(CustomPackage.indexJS())
```
cmgrote, line 258: Looks like we're doing this twice, back-to-back — copy/paste mistake (?)

pyatlan/pkg/templates/default_configmap.jinja2
```diff
@@ -0,0 +1,31 @@
...
+  config: |-
+    {
+      "properties": {
+{%- for key, value in pkg.ui_config.properties.items() %}
+        "{{ key }}": {
+          "type": "{{ value.type_ }}",
+          "required": {{ value.required | tojson}},
+    {%- if value.possible_values %}
+          "enum": [ {%- for val in value.possible_values.keys() %}{{ " " }}"{{ val }}"{%- if not loop.last %},{%- endif %}{% endfor -%}{{ " ]" }},
+          "enumNames": [ {%- for val in value.possible_values.values() %}{{ " " }}"{{ val }}"{%- if not loop.last %},{%- endif %}{% endfor -%}{{ " ]" }},
```
cmgrote, line 15: Just to confirm — `.values()` and `.keys()` return insertion order in both cases (they're lists, not sets)? Otherwise the keys and values will be skewed relative to each other and / or listed in the drop-down in a different order than what the user configured them in...

pyatlan/pkg/templates/logging_conf.jinja2
```diff
@@ -0,0 +1,48 @@
+[loggers]
```
cmgrote, line 1: Strange to me that this is a per-package template, when there's nothing templated in it — I guess this is to workaround something in the logging (?)

pyatlan/pkg/templates/package_config.jinja2
```diff
@@ -0,0 +1,87 @@
...
+
+class ConnectorAndConnection(BaseModel):
+    source: AtlanConnectorType
+    connections: list[str]
+
+
+def validate_connector_and_connection(v, values, config, field, **kwargs):
+    return ConnectorAndConnection.parse_raw(v)
+
+class CustomConfig(BaseModel):
```
cmgrote, line 43: Is this hard-coding the name of the (generated) custom package-specific config class to `CustomConfig` every time?

Would this be troublesome if we have a custom package that directly calls another custom package — each with their own (different) custom configs both named `CustomConfig`?

pyatlan/pkg/utils.py
```diff
@@ -0,0 +1,97 @@
...
+        self._env = Environment(  # noqa: S701
+            loader=PackageLoader("pyatlan.pkg", "templates")
+        )
+
+    def create_package(self):
+        self._root_dir.mkdir(parents=True, exist_ok=True)
+        with (self._root_dir / "index.js").open("w") as index:
+            index.write(CustomPackage.indexJS())
+        with (self._root_dir / "index.js").open("w") as index:
+            index.write(CustomPackage.indexJS())
```
cmgrote, line 38: Looks familiar... Don't we have this same class and method defined elsewhere already?

Can we dedup, or is this back to the classic Python inability to handle circular dependencies?

pyatlan/pkg/widgets.py
```diff
@@ -0,0 +1,1037 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2023 Atlan Pte. Ltd.
+import abc
+import json
+from dataclasses import dataclass, field
+
+# from dataclasses import dataclass
```
cmgrote, line 7: Let's drop these if they're truly unused? 🙏

pyatlan/pkg/widgets.py
```diff
@@ -0,0 +1,1037 @@
...
+    Field,
+    StrictBool,
+    StrictInt,
+    StrictStr,
+    dataclasses,
+    validate_arguments,
+)
+from pydantic.json import pydantic_encoder
+
+# from pydantic.dataclasses import dataclass
```
cmgrote, line 20: as above...

pyatlan/pkg/widgets.py
```diff
@@ -0,0 +1,1037 @@
...
+            grid=grid,
+        )
+        super().__init__(type_="string", required=required, ui=widget)
+
+
+@dataclasses.dataclass
+class MultipleUsersWidget(AbstractWidget):
+    def __init__(self, label: str, hidden: bool = False, help: str = "", grid: int = 8):
+        super().__init__(
+            widget="groupMultiple",
```
cmgrote, line 690: Recently fixed the same typo bug on Java SDK — this one should be `userMultiple`.

pyatlan/version.txt
```diff
@@ -1 +1 @@
-1.8.4
+1.8.5
```
cmgrote, line 1: Given the significant new functionality with the custom package toolkit (which we should call-out as experimental, still), suggest we bump to next minor release (1.9.0)

tests/unit/pkg/test_widgets.py
```diff
@@ -0,0 +1,1908 @@
...
+        sut = MultipleUsers(
+            label=LABEL,
+        )
+        assert sut.type_ == "string"
+        assert sut.required == IS_NOT_REQUIRED
+
+        ui = sut.ui
+        assert ui
+        assert isinstance(ui, MultipleUsersWidget)
+        assert ui.widget == "groupMultiple"
```
cmgrote, line 1215: `userMultiple` here, too

cmgrote (changes requested):
Few fixes and a few bits I'm unsure about (maybe logical reason we're doing them?)

pyatlan/pkg/models.py
ErnestoLoma: fixed

pyatlan/pkg/templates/default_configmap.jinja2
ErnestoLoma: Yes they are returned in insertion order.

pyatlan/pkg/templates/logging_conf.jinja2
ErnestoLoma: It's intended to be used by the custom package toolkit. I just made it a template in case we wanted to add any customization in the future.

pyatlan/pkg/templates/package_config.jinja2
ErnestoLoma: I'm not sure I'm following you. CustomConfig is the name of the class but it will in it's own .py file i.e. module so it will be namespaced with the module name so there shouldn't be a conflict.

ErnestoLoma: I would be interested on more info about a custom package calling another custom package. In that case wouldn't there be a CustomConfig that contains all the parameters for subsequent custom packages?

pyatlan/pkg/utils.py
ErnestoLoma: This was an oversight from some refactoring. This code was moved. I have deleted this.

pyatlan/pkg/widgets.py
ErnestoLoma: done

pyatlan/pkg/widgets.py
ErnestoLoma: Done

pyatlan/pkg/widgets.py
ErnestoLoma: done

pyatlan/version.txt
ErnestoLoma: done

tests/unit/pkg/test_widgets.py
ErnestoLoma: done

pyatlan/pkg/templates/package_config.jinja2
cmgrote: > I'm not sure I'm following you. CustomConfig is the name of the class but it will in it's own .py file i.e. module so it will be namespaced with the module name so there shouldn't be a conflict.

Ah, ok, that'll work.

> I would be interested on more info about a custom package calling another custom package. In that case wouldn't there be a CustomConfig that contains all the parameters for subsequent custom packages?

Yes — so let's say there is one custom package that is going to call 2 others in sequence: there'd be a total of 3 `CustomConfig` classes, but each namespaced separately. So sounds like it should be fine?

cmgrote (approved):
LGTM for an initial experimental version — once the tests are fixed, let's merge 👍

pyatlan/client/impersonate.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _get_client_info(self) -> ClientInfo:
+        client_id = os.getenv("CLIENT_ID")
+        client_secret = os.getenv("CLIENT_SECRET")
+        if not client_id or not client_secret:
+            raise ErrorCode.MISSING_CREDENTIALS.exception_with_parameters()
+        client_info = ClientInfo(client_id=client_id, client_secret=client_secret)
+        return client_info
+
+    def escolate(self) -> str:
```
Aryamanz29, line 77: A small typo here: `escolate` >> `escalate`

PR #231, opened 2024-01-24
Fixes parallel testing concurrency

Aryamanz29:
- Added [`concurrency`](https://docs.github.com/en/actions/using-jobs/using-concurrency) to the parallel integration tests. It will put workflow jobs in a waiting state if a similar integration test is already running in a workflow.

Screenshot 2024-01-24 at 8 05 20 PM


ErnestoLoma (approved):
Seems ok

PR #242, opened 2024-02-13
Migrate to `pydantic v2`

Aryamanz29:
Todos:

~~- [x] Refactor generation to create separate modules for each asset~~


cmgrote (approved):
Have reviewed most of the files (not the generated code in detail — relying on unit tests and integration tests to confirm that bit), and looks great! 👍 

Will get it merged into `main` so that we can do some thorough testing over the coming days prior to a release with all of these updates.

PR #248, opened 2024-02-21
DVX-265: Change all assets `create()` to `creator()` and `create_for_modification()` to `updater()`

Aryamanz29:
Todo:
~- [x] Currently, I have set [`DVX-263`](https://github.com/atlanhq/atlan-python/pull/247) as the base branch (to make it easier for us to see the changes in this PR). Once that is merged, I need to change the base branch back to `main`.~

### DIFF
--- a/pyatlan/generator/templates/methods/attribute/auth_policy.jinja2
+++ b/pyatlan/generator/templates/methods/attribute/auth_policy.jinja2
@@ -1,6 +1,6 @@
 
         @classmethod
-        def create(cls, name: str) -> {{ entity_def.name }}.Attributes:
+        def __create(cls, name: str) -> {{ entity_def.name }}.Attributes:
             validate_required_fields(["name"], [name])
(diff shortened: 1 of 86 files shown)


pyatlan/generator/templates/methods/attribute/auth_policy.jinja2
```diff
@@ -1,7 +1,6 @@
 
         @classmethod
-
-        def __create(cls, name: str) -> {{ entity_def.name }}.Attributes:
+        def create(cls, name: str) -> {{ entity_def.name }}.Attributes:
```
cmgrote, line 3: I think this was intentionally a private method to prevent anyone trying to use this internal mechanism directly (many things need to be set on a policy just-so in order for it to work as expected, which if someone calls this internal method directly they won't know about)

cmgrote (changes requested):
Just one check — we should leave this method private, if possible

PR #250, opened 2024-02-21
Fixes missing `NotFoundError`` exception in `AssetClient.find_._by_name` methods

### LINKED ISSUE #249: AtlanClient.asset.find_connections_by_name method does not return 'NotFoundError' when connection does not exist
## Describe the Bug

The `find_connections_by_name` method [here](https://github.com/atlanhq/atlan-python/blob/07c0ae1a4f039e6844acb668da0f338fa8c9a893/pyatlan/client/asset.py#L1321) returns an empty list instead of the `NotFoundError` as described in the doc string

## Steps to reproduce

NOTE: This expects the variables `atlan_base_url` and `atlan_api_key` to be set, not included here

```python

from pyatlan.errors import NotFoundError
from pyatlan.model.enums import AtlanConnectorType

# atlan_base_url = "myatlaninstance"
# atlan_api_key = "mykey"

atlan_client = AtlanClient(base_url=atlan_base_url, api_key=atlan_api_key)

try: 
  results = atlan_client.asset.find_connections("my_non_existent_connection",AtlanConnectorType.GCS)
  print(len(results)) 
except NotFoundError:
  print("Exception thrown, connection does not exist")
  
```

Instead of throwing the NotFoundError exception, it instead returns an empty list.

This could be expected behaviour if the doc string itself is wrong in the method

Aryamanz29:
Adds missing `NotFoundError` exception to the following methods:

- `find_personas_by_name`
- `find_purposes_by_name`
- `find_connections_by_name`

Fixes #249


pyatlan/client/asset.py
```diff
@@ -280,7 +280,9 @@ def find_personas_by_name(
             attributes=attributes,
         )
         results = self.search(search_request)
-        return [asset for asset in results if isinstance(asset, Persona)]
+        if personas := [asset for asset in results if isinstance(asset, Persona)]:
```
cmgrote, line 283: More curiosity than necessity, but are we not able to reuse the `_search_for_asset_with_name()` passing e.g. `Connection` / `Persona` / `Purpose` as the type and letting it handle the rest (i.e. not repeating code in all of these to run the search, list-comprehend the results, etc)?

cmgrote (approved):

pyatlan/client/asset.py
Aryamanz29: Sure, we can reuse `_search_for_asset_with_name()` method.

PR #253, opened 2024-02-26
DVX-233: Adds paging to `user`, `group`, and `membership` listings

Aryamanz29:


cmgrote (commented):
Looks good — only wondering if we should update (or add) an integration test that uses a very small page size (probably `1`) to ensure that the paging through results is as expected?

Since we don't want to rely on a consistent set of users in-tenant at a given point in time, you could just validate that with a page size of `1` we retrieve more than 1 unique result across pages.

tests/integration/admin_test.py
```diff
@@ -80,6 +81,89 @@ def test_retrieve_all_groups(client: AtlanClient, group: CreateGroupResponse):
...
+    assert current_page is not None
+    assert len(current_page) >= 1
+    for test_user in response:
+        assert test_user.username
+        assert test_user.email
+        assert test_user.attributes
+        assert test_user.login_events
+    current_page = response.current_page()
+    assert current_page is not None
+    assert len(current_page) == 0
```
cmgrote, line 142: Won't this still succeed if there is only a single user result across all pages?

May be difficult with our current setup for groups (we only have 1 group by default); but for users we should be checking that the total number is strictly greater than (`>`) 1, not gte (`>=`) I think.

Aryamanz29: Thanks @cmgrote!
For the user test, `response.total_record` should be `> 1`. I think we should also change `len(current_page) >= 1` to `len(current_page) == 1` in all pagination tests, given that the **limit** is set to **1** 👍

cmgrote (approved):

PR #257, opened 2024-02-29
DVX-269: Add task queue endpoint

Aryamanz29:
### Sample snippet:

```py

from pyatlan.model.fluent_tasks import FluentTasks
from pyatlan.model.task import TaskSearchRequest, AtlanTask, AtlanTaskStatus

client = AtlanClient()

task_request = (
    FluentTasks().where(AtlanTask.STATUS.match(AtlanTaskStatus.COMPLETE.value))
).to_request()

response = client.tasks.search(request=task_request)

for task in response:
	...
```

Docs:
Tags management
 - Monitor propagation


ErnestoLoma (approved):
Looks good

tests/integration/test_task_client.py
```diff
@@ -0,0 +1,122 @@
...
+        .page_size(1)
+        .sort(
+            by=SortItem(
+                field=AtlanTask.START_TIME.numeric_field_name,
+                order=SortOrder.DESCENDING,
+            )
+        )
+        .where(AtlanTask.ENTITY_GUID.eq(snowflake_column.guid))
+        .where(AtlanTask.TYPE.eq(AtlanTaskType.CLASSIFICATION_PROPAGATION_ADD.value))
+        .where(AtlanTask.STATUS.match(AtlanTaskStatus.PENDING.value))
```
cmgrote, line 71: Let's just keep an eye on this — my only concern here is that if we filter by `PENDING` there's a chance that the background task is actually picked up and run (and then no longer pending) before this test runs...

For repeatability, it might be safer to remove this pending condition — searching by entity GUID (will always be unique) alone may be sufficient, but GUID + the task type (classification_propagation_add) should be sufficient, I think?

cmgrote (changes requested):
Just one minor change that I think will improve test repeatability

tests/integration/test_task_client.py
Aryamanz29: Right! 
I've modified the search request to utilize the entity GUID along with the task type 👍

cmgrote (approved):

PR #262, opened 2024-03-07
DVX-280: Adds `Sigma` and `SQL` server crawler

Aryamanz29:
### Todos
- [x] Add tests
- [x] Update dev portal


pyatlan/model/packages/sigma_crawler.py
cmgrote, line 1: There are a number of spots in docstrings mentioning `sigma` — since it's a brand name, we should follow capitalization: `Sigma`

cmgrote (changes requested):
Just needs some docstring cleanup, I think 🙏

cmgrote (approved):

PR #268, opened 2024-03-14
Skip run query tests when when malformed connection was found

Aryamanz29:
ss


cmgrote (changes requested):
We need to figure out what changed and is causing these malformed connection errors in the first place — revising our code if needed to get the test working again, vs skipping the test entirely.

cmgrote:
I had reproduced the same error in the Java SDK, but since working through the changes leading up to this state am able to run this test successfully now there — so I believe we should be able to run it successfully now here, too.

Going to close this PR as a result.

PR #272, opened 2024-03-20
DVX-320: Adds `AtlanAPIResponse` model to support extra fields

Aryamanz29:


cmgrote (approved):
Cool — looks like all tests continue to pass, so let's do this 😁

PR #276, opened 2024-03-24
DVX-338: Adds `EnumDef` update method to add new `ElementDefs`

Aryamanz29:


pyatlan/cache/enum_cache.py
```diff
@@ -33,14 +33,17 @@ def refresh_cache(cls) -> None:
...
 
-        :param name: human-readable name of the enumeration
-        :returns: the enumeration definition
+        :param name: human-readable name of the enumeration.
+        :raises `NotFoundError`: if the enumeration with the given name does not exist.
+        :returns: enumeration definition
         """
-        return cls.get_cache()._get_by_name(name=name)
+        if not (enum := cls.get_cache()._get_by_name(name=name)):
+            raise ErrorCode.ENUM_NOT_FOUND.exception_with_parameters(name)
```
cmgrote, line 45: I'm OK with this, as I don't anticipate it'll be a heavily-used area. However, technically this is a breaking change in the interface. Let's just be sure that in the next release notes we indicate this change in behavior for this particular method as a breaking change?

ErnestoLoma (approved):
Looks fine

cmgrote (approved):
See comment related to next set of release notes — but looks good to merge.

pyatlan/cache/enum_cache.py
Aryamanz29: Noted!

PR #279, opened 2024-03-27
Bump to release `2.1.0`

Aryamanz29:


HISTORY.md
```diff
@@ -1,3 +1,20 @@
...
+### New features
+
+- Adds `EnumDef.update()` method to update an existing enumeration with more valid values.
+- Adds `get_by_name()` method to the `TypeDefClient` to retrieve a specific type definition from Atlan.
+- Adds `propagation_only_through_lineage` field to the `AtlanTag`, allowing users to specify that
+propagation will only occur downstream in the lineage and not within the hierarchy.
+
+### Breaking changes
+
+- `EnumCache.get_by_name()` method now raises `NotFoundError` if the enumeration with the given name does not exist.
```
cmgrote, line 12: Let's add a sentence to this explaining the previous behavior, e.g.:

> `EnumCache.get_by_name()` method now raises `NotFoundError` if the enumeration with the given name does not exist. Previously the method would only return `None` in case the enumeration did not exist, rather than raising an error.

cmgrote (changes requested):

PR #280, opened 2024-03-27
workflow api's for schedule queries

nishantmunjal7:
1. Add's two new endpoints from heracles to fetch schedule queries cron between a duration and missed schedule queries crons between a duration
2. Opens up a new function that can fetch details about schedule queries cron using saved query id
3. Re-triggers a workflow for schedule query, this is different from re-running a workflow. Re-running a workflow gets run using the user who is triggering it, but in case of schedule query, re-run has to happen from the owner of the workflow for queries to work.
4. Adds an new endpoint for updating the workflow owner, call to heracles

More info:
- https://www.notion.so/atlanhq/Python-SDK-Functions-for-Replaybility-ed63d0e992954b62a9d84edabded8612


Aryamanz29:
Hi @nishantmunjal7, thanks for your contribution! 🥳 🎉 

I've [pushed the refactored changes](https://github.com/atlanhq/atlan-python/pull/280/commits/bea4cb6768ba9eb2d6ec8af3fdc3bb6b7863e508) to this PR Could you please review these newly added methods manually again on the latest changes 🙏 and let us know if something needs to be updated 😄 

Currently, this feature is not enabled on our dev tenant (https://devx8.atlan.com), but we'll be adding integration for this feature soon. 💯

Aryamanz29 (approved):
I have resolved conflicts with the `main` branch. Everything looks good as of now, and this feature will be included in the upcoming release 🎉

PR #281, opened 2024-03-28
Fix DVX-352 Attempting to retrieve an MC Monitor asset by GUID is failing in field-sandbox

ErnestoLoma:
The problem was the code used to flatten structs was assuming the value in attributes for the struct_name always contained a list and `mcMonitorRuleScheduleConfig` contain a dict with the typeName and attributes.


Aryamanz29 (approved):
LGTM 🚀

PR #284, opened 2024-03-29
 DVX-359: Generate latest typedef models (`cube`, `cognite`, `spark`, `sparkjob` etc)

Aryamanz29:


pyatlan/generator/class_generator.py
cmgrote, line 1: Nice!

cmgrote (approved):

PR #287, opened 2024-04-02
DVX-330: Adds ability to update the set of connections, asset types, glossaries for custom metadata

Aryamanz29:


ErnestoLoma (approved):
Looks good

PR #288, opened 2024-04-02
 DVX-366: Add support for `restrict_propagation_through_hierarchy` in the `AtlanTag`

Aryamanz29:
- We previously assumed lineage-only propagation was added using `propagation_only_through_lineage`, but it's actually released with `restrict_propagation_through_hierarchy`.

- More info: https://atlanhq.atlassian.net/browse/PLT-1092


ErnestoLoma (approved):
Looks good

PR #291, opened 2024-04-03
Fixes field alias typo for `restrict_propagation_through_hierarchy` in the `AtlanTag` model.

Aryamanz29:
~Also updated the parameter name in tag methods from `restrict_propagation_through_hierarchy` to `restrict_hierarchy_propagation` for consistency with other parameters.~

### DIFF
--- a/pyatlan/model/core.py
+++ b/pyatlan/model/core.py
@@ -180,7 +180,7 @@ class AtlanTag(AtlanObject):
             "Whether to prevent this Atlan tag from propagating through "
             "hierarchy (True) or allow it to propagate through hierarchy (False)"
         ),
-        alias="restrictPropagationThroughHierachy",
+        alias="restrictPropagationThroughHierarchy",
     )
     validity_periods: Optional[List[str]] = Field(default=None, alias="validityPeriods")
     _source_tag_attachements: List[SourceTagAttachment] = PrivateAttr(


pyatlan/client/asset.py
```diff
@@ -742,7 +742,7 @@ def _modify_tags(
         propagate: bool = True,
         remove_propagation_on_delete: bool = True,
         restrict_lineage_propagation: bool = True,
-        restrict_propagation_through_hierarchy: bool = False,
+        restrict_hierarchy_propagation: bool = False,
```
ErnestoLoma, line 745: Why the name change?

ErnestoLoma (changes requested):

pyatlan/client/asset.py
Aryamanz29: I've updated the parameter name in tag methods from `restrict_propagation_through_hierarchy` to `restrict_hierarchy_propagation` to maintain consistency with other parameters (e.g., `restrict_lineage_propagation`).

Although this change results in a backward incompatible change, we can mention it as **breaking change** in release `2.1.3`.

ErnestoLoma: I think we should be using the same name as on the tag. Personally I think restrict_lineage_propagation so match the value in AtlanTag, but I wouldn't change it now. And since we already did a release with restrict_propagation_through_hierarchy I think we should keep it the same.

Aryamanz29: Okay, cool!

I'll revert that 👍

ErnestoLoma: Also it seems to me if you want to only propagate through the hierarchy then it's unclear to me whether you would set restrict_hierarchy_propagation to True or False. It seems to me that restrict_hierarchy_propagation True could mean don't propagate to the hierarchy. On the other hand I can understand wanting consistency and since the properties on AtlanTag will probably be rarely used by the users, I guess consistency is more important.

ErnestoLoma (approved):

PR #296, opened 2024-04-15
Allow newer versions of networkx

fox:
Bumps `networkx`  to >= 3.1 since `pyatlan` seem to work fine with newer versions (`networkx` 3.2 and 3.3).

Having version fixed at 3.1. can cause conflicts with other packages that depend on newer versions of `networkx`. This PR resolves such conflicts.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 requests>=2.24
 pydantic~=2.6.1
 jinja2==3.1.3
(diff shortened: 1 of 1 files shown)


Aryamanz29:
Thanks @fox!
Superseded by: https://github.com/atlanhq/atlan-python/pull/298

PR #299, opened 2024-04-16
 DVX-387: Corrects `qualified_name` population for related entities in glossary objects

Aryamanz29:
- Also added `relation_attributes=['name']` to the `IndexSearchRequest` to include the related asset name.

### DIFF
--- a/pyatlan/generator/templates/methods/asset/atlas_glossary.jinja2
+++ b/pyatlan/generator/templates/methods/asset/atlas_glossary.jinja2
@@ -1,18 +1,17 @@
 
     @root_validator()
     def _set_qualified_name_fallback(cls, values):
+        guid = values.get("guid")
         attributes = values.get("attributes")
         unique_attributes = values.get("unique_attributes")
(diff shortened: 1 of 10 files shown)


pyatlan/generator/templates/methods/asset/atlas_glossary.jinja2
```diff
@@ -1,8 +1,18 @@
 
     @root_validator()
     def _set_qualified_name_fallback(cls, values):
```
ErnestoLoma, line 3: The code in these methods looks like it's duplicated. I would refactor it into a function.

ErnestoLoma (changes requested):

pyatlan/generator/templates/methods/asset/atlas_glossary.jinja2
Aryamanz29: I have removed the code duplication from these methods, so now they look clean. I'll leave it as is. Thanks!

Aryamanz29:
Build is ✅
Merging this now, as it will be included in today's release.

PR #301, opened 2024-04-17
  DVX-392: Fixes explicit assignment of `None` to asset attributes in `BulkRequest`

Aryamanz29:
- Also fixes issues with multiple remove/append relationships
Previously, values were being overwritten inside append/remove dictionaries.
```py
# Test replace and append (list) with multiple relationships
 term1.attributes.see_also = [
            AtlasGlossaryTerm.ref_by_guid(guid=term2.guid),
            AtlasGlossaryTerm.ref_by_guid(
                guid=term3.guid, semantic=SaveSemantic.APPEND
            ),
        ]
 term1.attributes.preferred_to_terms = [
            AtlasGlossaryTerm.ref_by_guid(
                guid=term3.guid, semantic=SaveSemantic.APPEND
            ),
        ]
```

### DIFF
--- a/pyatlan/model/core.py
+++ b/pyatlan/model/core.py
@@ -277,10 +277,10 @@ class BulkRequest(AtlanObject, GenericModel, Generic[T]):
 
         # Initialize set for attributes to exclude from serialization
         exclude_attributes = set()
-        # Manually need to set these to "None" so that we can exclude
+        # Manually need to set these to "{}" so that we can exclude
         # them from the request playload when they're not set by the user
-        asset.remove_relationship_attributes = None
-        asset.append_relationship_attributes = None
+        asset.remove_relationship_attributes = {}
+        asset.append_relationship_attributes = {}
         # Process relationship attributes and update exclusion set
         for attribute in asset.attributes:
             exclude_attributes.update(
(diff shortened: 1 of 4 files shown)


pyatlan/model/core.py
```diff
@@ -275,17 +275,39 @@ def process_attributes_and_flush_cm(cls, asset):
...
+        # https://docs.pydantic.dev/1.10/usage/exporting_models/#advanced-include-and-exclude
+        exclude_relationship_attributes = {
+            key: True
+            for key in [
+                "remove_relationship_attributes",
+                "append_relationship_attributes",
+            ]
+            if not getattr(asset, key)
+        }
+        if exclude_attributes:
```
ErnestoLoma, line 299: Regarding my previous comment I added the following line and all the integration and unit tests passed so it would seem that you only need to check the attributes in `asset.attributes.__fields_set__`:
`assert exclude_attributes.issubset(asset.attributes.__fields_set__)`

pyatlan/model/core.py
```diff
@@ -275,17 +275,39 @@ def process_attributes_and_flush_cm(cls, asset):
...
+        # Initialize set for attributes to exclude from serialization
+        exclude_attributes = set()
+        # Manually need to set these to "{}" so that we can exclude
         # them from the request playload when they're not set by the user
-        asset.remove_relationship_attributes = None
-        asset.append_relationship_attributes = None
+        asset.remove_relationship_attributes = {}
+        asset.append_relationship_attributes = {}
+        # Process relationship attributes and update exclusion set
         for attribute in asset.attributes:
```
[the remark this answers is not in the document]
Aryamanz29: Yes, I forgot about this internal attribute `(__fields_set__)`, which will further reduce iterations to only process user-set attributes. Thanks!

pyatlan/model/core.py
Aryamanz29: Yes, it works!
Thanks.

ErnestoLoma (approved):
Looks good

PR #318, opened 2024-05-08
Initial implementation of DVX-430

ErnestoLoma:
Need to add an additional parameter to allow the specification of attributes to be included in related assets that may be returned by the hierarchy search.


Aryamanz29 (approved):
LGTM 🚀

PR #320, opened 2024-05-09
DVX-432: Utilize generated `atlas core` enums in the template

Aryamanz29:
When a new `atlas_core` enum is introduced and generated models depend on those enums, a `NameError` occurs due to the hardcoding of [enums in the template](https://github.com/atlanhq/atlan-python/blob/main/pyatlan/generator/templates/imports.jinja2#L18). This issue arises because the hardcoded enums are not up-to-date with the latest enums. To address this problem, I inject the atlas enums directly into the template. However, I retain the hardcoded enums import statement in the template, as it contains all enums (both `pyatlan` defined and `generated enums`). After generation, pre-commit formatting checks are used to clean up any duplicate or unused import statements.

### DIFF
--- a/pyatlan/generator/class_generator.py
+++ b/pyatlan/generator/class_generator.py
@@ -635,12 +635,13 @@ class Generator:
     def render_modules(self, modules: List[ModuleInfo]):
         self.render_init(modules)  # type: ignore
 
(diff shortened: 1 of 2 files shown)


ErnestoLoma (approved):
Looks good

PR #324, opened 2024-05-14
Bump to release `2.1.9`

Aryamanz29:


HISTORY.md
```diff
@@ -1,3 +1,14 @@
+## 2.1.9 (May 14, 2024)
+
+### New features
+
+- Added support for `BigQuery`, `DynamoDB` and `Postgres` crawlers.
+- Added logic to `Referenceable` model to determine the correct sub type.
```
ErnestoLoma, line 6: I would put this under bug fix

ErnestoLoma (changes requested):
One minor change

HISTORY.md
Aryamanz29: Correct, my bad.
I'll update that 👍

ErnestoLoma (approved):

PR #325, opened 2024-05-14
Adds `pyatlan-formatter`

Aryamanz29:
While working on: https://atlanhq.atlassian.net/browse/DVX-433, I observed that the [pre-commit script only executes on Git-tracked files](https://pre-commit.com/#4-optional-run-against-all-the-files). However, this poses an issue when we generate new modules and attempt to run the formatter on them, as it requires tracking (staging) those files before running `pre-commit run --all-files`. This adds unnecessary steps for users who may not be familiar with pre-commit. To address this, I propose the addition of a simple shell script, `./pyatlan-formatter`, which can handle formatting tasks for both tracked and untracked files.


pyatlan/generator/class_generator.py
```diff
@@ -642,6 +643,7 @@ def render_module(self, asset_info: AssetInfo, enum_defs: List["EnumDefInfo"]):
                 "asset_info": asset_info,
                 "existz": os.path.exists,
                 "enum_defs": enum_defs,
+                "templates_path": TEMPLATES_DIR.absolute().as_posix(),
```
ErnestoLoma, line 646: Curious why was this change needed?

pyatlan/generator/templates/module.jinja2
```diff
@@ -40,7 +40,7 @@ class {{ entity_def.name }}({{super_classes[0]}} {%- if "Asset" in super_classes
     {% include 'referenceable_attributes.jinja2' %}
 {%- else %}
     {% set file_name = 'methods/asset/' + entity_def.name | to_snake_case + '.jinja2' %}
-    {%  if existz('templates/' + file_name) %}
+    {%  if existz(templates_path + '/' + file_name) %}
```
ErnestoLoma, line 43: Why the change?

ErnestoLoma (changes requested):
Just a couple of questions

pyatlan/generator/class_generator.py
Aryamanz29: Related to: https://github.com/atlanhq/atlan-developer-portal/pull/300

I want to run generator scripts from the project root directory, for example: `python3 pyatlan/generator/class_generator.py`. In this case, the hardcoded path in the templates is not respected. Therefore, we should use the `pathlib` path in the templates.

pyatlan/generator/templates/module.jinja2
Aryamanz29: Related to: https://github.com/atlanhq/atlan-developer-portal/pull/300

I want to run generator scripts from the project root directory, for example: `python3 pyatlan/generator/class_generator.py`. In this case, the hardcoded path in the templates is not respected. Therefore, we should use the `pathlib` path in the templates.

PR #326, opened 2024-05-15
Fixes typo in `IAM` role methods of `DynamoDB` and `PostgreSQL` crawlers

Aryamanz29:
I found this while working on: https://github.com/atlanhq/atlan-developer-portal/pull/301. This will result in a [breaking change](https://github.com/atlanhq/atlan-python/pull/327). Moving forward, I'll aim to merge docs PRs first and then cut the release here to prevent such issues in the future.


ErnestoLoma (commented):
Looks fine

PR #331, opened 2024-05-21
DVX-444 Added `find_by_name` methods for domain and products 

Aryamanz29:
- Fixed data domain and product `creator()` methods.
- Updated data mesh integration tests.
- Generated latest typedef models


ErnestoLoma (commented):
Looks good

Issue #333, opened 2024-05-21
Get more flexibility on the Pydantic dependency


josecsotomorales:
We are using Atlan Python SDK to integrate with our platform, we have an issue with conflicting versions of Pydantic as we usually run the latest version, currently `2.7.1`. Can you change it from your `requirements` to 2.x.x?

https://github.com/atlanhq/atlan-python/blob/beea7e2c20c62e3ee4862f6dbdc7be51854c172b/requirements.txt#L2

Thanks!

PR #335, opened 2024-05-26
Updated the `pydantic` dependency to provide more flexibility for users

Aryamanz29:
Related to #333


ErnestoLoma (changes requested):
Did you run the integration tests with later versions of pydantic?

Aryamanz29:
> Did you run the integration tests with later versions of Pydantic?

@ErnestoLoma I have tested on `pydantic-2.7.1` (latest) locally, and the [GitHub CI also passes](https://github.com/atlanhq/atlan-python/actions/runs/9243652107/job/25428528240?pr=335). I hope we don't face any issues with `pydantic>=2.0.0,<3.0.0` in the future, as long as Pydantic follows semantic versioning 🤞
It's also worth noting that in `pyatlan`, we still don't use any Pydantic v2 features (we are still using the v1 compatibility layer `pydantic.v1` in our models), so I think we're good to go 😄

ErnestoLoma (approved):

PR #337, opened 2024-05-28
DVX-486 Update trim_to_required on AtlasGlossaryCategory to use updater instead of create_for_modifications

ErnestoLoma:


Aryamanz29 (approved):
LGTM 🚀

PR #341, opened 2024-05-30
DVX-481: Added support for handling data domains and products in custom metadata

Aryamanz29:


pyatlan/model/typedef.py
```diff
@@ -639,6 +657,31 @@ def applicable_glossary_types(self, glossary_types: Set[str]):
...
+        Data product type names to which to restrict the attribute.
+        These cover asset types in data products and data domains.
+        Only assets of one of these types will have this attribute available.
+        """
+        if self.options and self.options.applicable_domain_types:
+            return set(json.loads(self.options.applicable_domain_types))
+        return set()
+
+    @applicable_domain_types.setter
+    def applicable_domain_types(self, domain_types: Set[str]):
```
ErnestoLoma, line 672: I think the type should be Set[Literal["DataDomain", "DataProduct"]] not Set[str]

pyatlan/model/typedef.py
```diff
@@ -717,6 +781,8 @@ def create(
         applicable_glossaries: Optional[Set[str]] = None,
         applicable_glossary_types: Optional[Set[str]] = None,
         applicable_other_asset_types: Optional[Set[str]] = None,
+        applicable_domains: Optional[Set[str]] = None,
+        applicable_domain_types: Optional[Set[str]] = None,
```
ErnestoLoma, line 785: I think the type should be Optional[Set[Literal["DataDomain", "DataProduct"]]] not Optional[Set[str]]

ErnestoLoma (changes requested):
Couple of small changes.

pyatlan/model/typedef.py
Aryamanz29: Thanks!

PR #343, opened 2024-05-31
 DVX-390: Overloaded creator methods to explicitly handle ancestral `qualifiedName` and `name`

Aryamanz29:
- Added overloaded `creator()` methods to explicitly manage ancestral `qualifiedName` and `name`.
- Ensured backward compatibility by maintaining existing method signatures.

## TODO

- [x] Add overloaded methods to other assets.
- [x] Update/Add unit and integration tests.


Aryamanz29:
Java SDK implementation: https://github.com/atlanhq/atlan-java/pull/642

pyatlan/model/assets/a_d_l_s_object.py
```diff
@@ -34,20 +34,45 @@
...
+    ) -> ADLSObject: ...
+
+    @overload
+    @classmethod
+    def creator(
+        cls,
+        *,
+        name: str,
+        adls_container_qualified_name: str,
+    ) -> ADLSObject: ...
```
ErnestoLoma, line 55: I think the most common method should be first since that seems to be the order in the IDE

pyatlan/model/assets/airflow_task.py
```diff
@@ -23,13 +23,33 @@
...
+
+    @overload
+    @classmethod
+    def creator(
+        cls,
+        *,
+        name: str,
+        airflow_dag_qualified_name: str,
+    ) -> AirflowTask: ...
+
```
ErnestoLoma, line 44: Again I think the most common method should be first

pyatlan/model/assets/column.py
```diff
@@ -27,24 +27,79 @@
...
+    @overload
+    @classmethod
+    def creator(
+        cls,
+        *,
+        name: str,
+        parent_qualified_name: str,
+        parent_type: type,
+        order: int,
+    ) -> Column: ...
```
ErnestoLoma, line 58: Again most common/simplest method first

pyatlan/model/assets/column.py
```diff
@@ -1354,48 +1409,87 @@ class Attributes(SQL.Attributes):
         @classmethod
         @init_guid
         def create(
```
ErnestoLoma, line 1411: I thought the create was deprecated, if so, I don't think we should modify it.

pyatlan/model/assets/g_c_s_object.py
```diff
@@ -25,14 +25,41 @@
...
+        cls,
+        *,
+        name: str,
+        gcs_bucket_qualified_name: str,
+        connection_qualified_name: str,
+    ) -> GCSObject: ...
+
+    @overload
+    @classmethod
+    def creator(
```
ErnestoLoma, line 40: again most common first

pyatlan/model/assets/g_c_s_object.py
```diff
@@ -416,18 +443,28 @@ class Attributes(GCS.Attributes):
         @classmethod
         @init_guid
         def create(
```
ErnestoLoma, line 445: again I don't think we should modify deprecated methods.

ErnestoLoma (changes requested):
I think the most common/simplest overridden method should appear first since that determines how it appears in the IDE.
I don't think we should make any changes to the create method since it is deprecated.

pyatlan/model/assets/g_c_s_object.py
Aryamanz29: Its `GCSObject.Attributes.create()` method.

pyatlan/generator/templates/methods/asset/a_d_l_s_container.jinja2
```diff
@@ -1,12 +1,40 @@
...
+    ) -> ADLSContainer: ...
+
+    @overload
+    @classmethod
+    def creator(
+        cls,
+        *,
+        name: str,
+        adls_account_qualified_name: str,
+        connection_qualified_name: Optional[str] = None,
```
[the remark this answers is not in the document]
Aryamanz29: I think you've reviewed an `outdated` file; it's absent in the latest commit.

Please see: https://github.com/atlanhq/atlan-python/blob/DVX-390/pyatlan/generator/templates/methods/asset/a_d_l_s_container.jinja2#L12-L18

ErnestoLoma (approved):

PR #352, opened 2024-06-19
Bump to release `2.3.0`

Aryamanz29:


HISTORY.md
```diff
@@ -1,3 +1,21 @@
+## 2.3.0 (June 19, 2024)
+
+### Breaking changes
+
+- Introduced a new pagination approach in `AssetClient.search()` and `FluentSearch.execute()` called **bulk search** (disabled by default). It minimizes system impact when handling large result sets. The SDK switches to this search operation automatically if results exceed a predefined threshold (i.e: `100,000` results). Alternatively, users can enable bulk search explicitly by setting `bulk=True` in `AssetClient.search()` or `FluentSearch.execute()`. One side effect of this search operation is that any requested sorting of the results will be overridden. If you need to sort such a large number of results, you will have to do so after retrieving them.
```
cmgrote, line 5: Let's change this:

> One side effect of this search operation is that any requested sorting of the results will be overridden. If you need to sort such a large number of results, you will have to do so after retrieving them.

To:

> The breaking change is in regards to searches that return more than `100,000` results.

pyatlan/client/asset.py
```diff
@@ -1848,14 +1876,18 @@ def _get_next_page(self):
...
+        is_bulk_search = (
+            self._bulk or self._approximate_count > self._MASS_EXTRACT_THRESHOLD
+        )
 
-        if self._bulk:
+        if is_bulk_search:
             LOGGER.debug(
-                "Bulk search option is enabled. Ignoring requests for default offset-based "
-                "paging and switching to a creation timestamp-based paging approach."
+                self._get_bulk_search_log_message(),
```
cmgrote, line 1885: I'm not sure we should log this on every single page — seems unnecessary (once the decision on approach is made for the first page, it'll remain the same for all subsequent pages, right?) — can we log this only once (up-front) when the decision on paging approach is made rather than on every single page (will flood logs with repetitive messages)?

cmgrote (changes requested):
Couple minor changes

PR #358, opened 2024-07-15
CXD-114 : Add Prefect etl connector in enum

prateekrai-atlan:
We have added Prefect frontend configs in ETL for connection and process. Extending the functions for python SDK.


Aryamanz29 (approved):
LGTM 🚀

PR #361, opened 2024-07-19
CXD-91 : Added Superset creator methods for supported assets

prateekrai-atlan:
We have added Superset models as typedefs and have created frontend configs for the same. Extending this for python SDK support with creator methods for each new asset type. Superset can be compared with Preset with only the workspace asset being a differentiator between the two.


Aryamanz29 (approved):
LGTM 🚀

PR #372, opened 2024-08-14
DVX-528 Commit final version of code to deploy to anaconda

ErnestoLoma:


Aryamanz29 (changes requested):
LGTM ✅ 
just one doubt below:

.github/workflows/build_and_upload_conda_packages.yaml
```diff
@@ -10,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
Aryamanz29, line 15: I think we should add **Python 3.8** here as well, since `pyatlan` supports [python>=3.8.](https://github.com/atlanhq/atlan-python/blob/main/setup.py#L71)

ErnestoLoma: Will do.

ErnestoLoma: done.

Aryamanz29 (approved):
LGTM ✅️

PR #373, opened 2024-08-26
DVX-591: Fixed interpretation of `text-only` indexed fields

Aryamanz29:
- Generated latest typedefs models.

Similar to: https://github.com/atlanhq/atlan-java/pull/762/


ErnestoLoma (approved):
Looks good

Aryamanz29:
The `lineage_test` [failure](https://github.com/atlanhq/atlan-python/actions/runs/10557844444/job/29246287970) is unrelated to this change, as it also fails on the `main` branch:
ss

PR #416, opened 2024-11-13
CJ-181 : `Application Container` Creator and Tests (new typedef)

prateekrai-atlan:
# 🆕 
New Asset Typedef called `Application` with one asset called `Application Container`. Details and Need : 
- Application has a shared attribute called `applicationId` which is a string type and stores the ID of the application in the `Application Container`.
- Application Container has a ONE-TO-MANY relationship with any asset under Catalog. Thus any asset under Catalog can have only one application container as parent, but application container has many child catalog assets.
- A de-normalized attribute called `assetApplicationQualifiedName` to store the qualified name of the parent application container at the catalog level. All catalog assets will inherit this attribute.

# 🧪 
Integration Test include : 
- Creator Test
- Overload Creator Test
- CRUD operation Test


Aryamanz29:
Re-ran the failed tests on https://fieldtec28.atlan.com locally, and they passed ✅. 
Merging this PR now 💯

PR #428, opened 2024-11-22
FT-773: Add X-Atlan-Client-Origin header to session requests

pavanmanish-atlan:


Aryamanz29 (changes requested):
Looks good!  
Here are a few minor requested changes:  
- I would suggest adding a unit test in [test_client.py](https://github.com/atlanhq/atlan-python/blob/main/tests/unit/test_client.py) to assert that the `AtlanClient` instance includes these headers. You can refer to [this example](https://github.com/atlanhq/atlan-python/blob/main/tests/integration/custom_package_test.py#L96) for guidance on asserting expected headers for custom packages.

pyatlan/client/atlan.py
```diff
@@ -133,6 +133,7 @@ def get_session():
             "x-atlan-agent": "sdk",
             "x-atlan-agent-id": "python",
             "User-Agent": f"Atlan-PythonSDK/{VERSION}",
+            "X-Atlan-Client-Origin": "product_sdk",
```
Aryamanz29, line 136: Let's add the custom header above `User-Agent`. I would also change it to lowercase to ensure consistency with other custom headers, as [HTTP headers are case-insensitive:](https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive):

```suggestion
            "x-atlan-client-origin": "product_sdk",
            "User-Agent": f"Atlan-PythonSDK/{VERSION}",
```

Aryamanz29 (commented):
Since there are [some checks failing](https://github.com/atlanhq/atlan-python/actions/runs/11968677664/job/33368056884?pr=428) please fix those issues.

You can use the following commands locally before pushing your commit:


1. For code formatting: 🧹 

   ```
   ./pyatlan-formatter
   ```

2. For code QA checks (`black`, `flake8`, `mypy` checks): ✅ 

   ```
   ./qa-checks
   ```

tests/unit/test_client.py
```diff
@@ -2204,3 +2205,16 @@ def test_asset_attribute_none_assignment(self):
...
+    expected = {
+        "User-Agent": f"Atlan-PythonSDK/{VERSION}",
+        "Accept-Encoding": "gzip, deflate",
+        "Accept": "*/*",
+        "Connection": "keep-alive",
+        "x-atlan-agent": "sdk",
+        "x-atlan-agent-id": "python",
+        "x-atlan-client-origin": "product_sdk",
+    }
+    assert expected == client._session.headers
```
Aryamanz29, line 2220: Missing newline here ^

Aryamanz29: https://github.com/atlanhq/atlan-python/pull/428#pullrequestreview-2453575421

Aryamanz29 (approved):
LGTM 🚀

PR #431, opened 2024-11-29
FT-814:Add projection support for groups endpoint

vaibhavatlan:
- Added Columns project support inside of groups.get_all() function. 
- Added roles in the AtlanGroup Class as it was missing. (additional change)


pyatlan/client/constants.py
```diff
@@ -15,6 +15,7 @@
 
 ROLE_API = "roles"
 GROUP_API = "groups"
```
ErnestoLoma, line 17: The v2/groups endpoint doesn't replace all previous occurrences of groups? It's only used when retrieving groups?

ErnestoLoma (commented):
Just a couple of questions.

pyatlan/client/group.py
```diff
@@ -96,8 +96,9 @@ def get(
         limit: Optional[int] = 20,
         post_filter: Optional[str] = None,
         sort: Optional[str] = None,
-        count: bool = True,
+        count: bool = False,
```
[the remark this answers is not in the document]
vaibhavatlan: This was changed to False because we are not retrieving the number of filtered records or the total number of records in our result response. Instead, we are directly outputting the records. Since the default value for this setting is True, it would have impacted the test cases, leading to inconsistencies. Changing it to False ensures the expected behavior aligns with our current implementation.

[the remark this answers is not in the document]
ErnestoLoma: I'm not following. I see that we are populating `total_records` attribute of `GroupResponse` from the `totalRecord` field in the returned JSON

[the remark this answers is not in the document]
vaibhavatlan: My bad. Apologies for the confusion earlier. I've reverted it back to True. Thanks for pointing that out!

ErnestoLoma (approved):
LGTM

tests/unit/test_client.py
```diff
@@ -2219,3 +2224,72 @@ def test_atlan_client_headers(client: AtlanClient):
...
+
+    columns = ["alias"]
+    groups = group_client.get_all(limit=10, columns=columns)
+
+    assert len(groups) == 2
+    assert groups[0].id == "1"
+    assert groups[0].alias == "Group1"
+    mock_api_caller._call_api.assert_called_once()
+    query_params = mock_api_caller._call_api.call_args.kwargs["query_params"]
+    assert query_params["columns"] == columns
```
Aryamanz29, line 2275: I would place `mock_api_caller.reset_mock()` at the end of my unit test assertions instead of at the start.
```suggestion
    mock_page_1 = [
        {"id": "1", "alias": "Group1"},
        {"id": "2", "alias": "Group2"},
    ]
    mock_api_caller._call_api.side_effect = [
        {"records": mock_page_1},
    ]

    columns = ["alias"]
    groups = group_client.get_all(limit=10, columns=columns)

    assert len(groups) == 2
    assert groups[0].id == "1"
    assert groups[0].alias == "Group1"
    mock_api_caller._call_api.assert_called_once()
    query_params = mock_api_caller._call_api.call_args.kwargs["query_params"]
    assert query_params["columns"] == columns
    mock_api_caller.reset_mock()
```

tests/unit/test_client.py
```diff
@@ -2219,3 +2224,72 @@ def test_atlan_client_headers(client: AtlanClient):
...
+
+
+def test_get_all_sorting(group_client, mock_api_caller):
+    mock_api_caller.reset_mock()
```
Aryamanz29, line 2279: I would place `mock_api_caller.reset_mock()` at the end of my unit test assertions instead of at the start.

Aryamanz29 (approved):
LGTM
Just one minor requested change:

tests/unit/test_client.py
vaibhavatlan: Done!

PR #434, opened 2024-12-04
FT-831: Updated `CredentialListResponse` fields and handled `None` re…

Aryamanz29:
- Renamed `CredentialResponseList` to `CredentialListResponse` to ensure consistent response model naming.  
- Updated handling for `records: null` in the response to populate the model with an empty list `[]` instead of `None`, which is non-iterable.  
- Updated `level` and `connection` fields to use `Optional[Union[Dict[str, Any], str]]`, as they can be strings, preventing Pydantic validation errors:

```json
   "level": "user",
   "connection": "default/bigquery/1697545730"
```


vaibhavatlan (approved):
LGTM 🚀

PR #436, opened 2024-12-05
Updated integration test to align to new tenant setup

Aryamanz29:


ErnestoLoma (commented):
What about test_workflow_client.test_get_all_credentials_with_limit_and_offset that was failing because there's only one credential in the tenant?
Is this for FT-825?

Aryamanz29:
>What about test_workflow_client.test_get_all_credentials_with_limit_and_offset that was failing because there's only one credential in the tenant?
Is this for FT-825?

I’m not aware of the failing test and [FT-825 ticket](https://atlanhq.atlassian.net/browse/FT-825): `test_get_all_credentials_with_limit_and_offset` 😅 
I was only fixing the integration tests that [failed yesterday ](https://github.com/atlanhq/atlan-python/actions/runs/12162058149)during the release. I think we can remove this test as it seems to be a duplicate (cc: @vaibhavatlan what do you think?). Since there’s already an [integration test (`test_get_all_credentials_with_filter_limit_offset`)](https://github.com/atlanhq/atlan-python/pull/436/files#diff-f240d66cbdc44a7ab6c23a713037e3e6dcc4b640f16c4eff97f6ce9f9b06b7afR283-R294) that covers the same scenario with `snowflake` credentials].

vaibhavatlan:
> > What about test_workflow_client.test_get_all_credentials_with_limit_and_offset that was failing because there's only one credential in the tenant?
> > Is this for FT-825?
> 
> I’m not aware of the failing test and [FT-825 ticket](https://atlanhq.atlassian.net/browse/FT-825): `test_get_all_credentials_with_limit_and_offset` 😅 I was only fixing the integration tests that [failed yesterday ](https://github.com/atlanhq/atlan-python/actions/runs/12162058149)during the release. I think we can remove this test as it seems to be a duplicate (cc: @vaibhavatlan what do you think?). Since there’s already an [integration test (`test_get_all_credentials_with_filter_limit_offset`)](https://github.com/atlanhq/atlan-python/pull/436/files#diff-f240d66cbdc44a7ab6c23a713037e3e6dcc4b640f16c4eff97f6ce9f9b06b7afR283-R294) that covers the same scenario with `snowflake` credentials].

Yes, you’re absolutely right. `test_get_all_credentials_with_limit_and_offset` addresses the same scenario as `test_get_all_credentials_with_filter_limit_offset` for Snowflake credentials. The failure occurs because, as Ernest pointed out, the new tenant `fieldtec33` only has one credential available. Since the integration test `test_get_all_credentials_with_filter_limit_offset` already covers this case, we can safely remove the failing test.

ErnestoLoma (approved):
LGTM

PR #441, opened 2024-12-10
FT-818: Implement query rewrite for large AuditSearch responses

vaibhavatlan:
Implementing a query rewrite for a time-based pagination approach for search, known as bulk search, based on asset creation timestamps. This approach is used when the number of results exceeds the predefined threshold (i.e: 100,000 assets). Users can explicitly run this search by setting the optional keyword argument bulk=True in the FluentSearch.execute() or asset.search() methods.

Example Code:
```
from pyatlan.model.audit import AuditSearchRequest
from pyatlan.model.assets import AtlasGlossaryTerm
from pyatlan.model.search import Bool,Term
from pyatlan.model.fluent_search import FluentSearch,DSL

client = AtlanClient()
guid="2be6f58d-fd8a-44fe-a43e-f97b150cd1dd"
dsl = DSL(
        query=Bool(filter=[Term(field="entityId", value=guid)]),
        sort=[],
        size=10,
        from_=0,
    )
audit_search_request = AuditSearchRequest(dsl=dsl)
response = client.audit.search(criteria=audit_search_request,bulk=True)

for result in response:  
    action = result.action
    user = result.user  
    when = result.timestamp 
    detail = result.detail.user_description  
    print(action,user,when,detail)
    print()
```


Aryamanz29 (changes requested):
Hey @vaibhavatlan, nice progress!

A few things we need to do next:
- Have you tested the new paging approach locally? Is it working as expected?
- Can you start by adding some `unit` tests first (we’ll add integration tests later)?

You can take inspiration from this PR (`time-based` paging approach for `index search`) for writing `unit` tests: https://github.com/atlanhq/atlan-python/pull/351

Let’s focus on **(1) and (2)^^** for now. I'll come back to review this PR in detail once I finish reviewing the open ones!

vaibhavatlan:
I have done local testing for it and it is working as expected.
I have added unit tests for it as well.

Aryamanz29 (approved):
LGTM!

I found another edge case with the timestamp-based pagination approach. I'll address it in a separate PR: https://atlanhq.atlassian.net/browse/FT-865

PR #444, opened 2024-12-11
FT-790: Remove possibility of overrunning search pages when total results are just under the threshold

vaibhavatlan:
>  It’s currently possible to hit an edge case where e.g. the total number of results is 9932, but because we page by default at 300 per page we end up sending a request from offset 9900 with a size of 300 <— in total this would take us above the 10000 threshold (to 10200) <— which Elastic will throw an error on.

We should ensure that we properly mitigate this edge case.

Remove possibility of overrunning search pages when total results are just under the threshold:
 - DEFAULT_PAGE_SIZE = 300
  - _MASS_EXTRACT_THRESHOLD = 100000 - DEFAULT_PAGE_SIZE

### DIFF
--- a/pyatlan/client/asset.py
+++ b/pyatlan/client/asset.py
@@ -1793,8 +1793,9 @@ class IndexSearchResults(SearchResults, Iterable):
     query.
     """
 
-    DEFAULT_PAGE_SIZE = 300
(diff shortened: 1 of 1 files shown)


Aryamanz29 (approved):
LGTM 🚀

PR #449, opened 2024-12-19
FT-853: Implement query rewrite for Searchlog responses

vaibhavatlan:
Implementing a query rewrite for a time-based pagination approach for search, known as bulk search, based on asset creation timestamps. This approach is used when the number of results exceeds the predefined threshold (i.e: 10,000 assets). Users can explicitly run this search by setting the optional keyword argument bulk=True in the  search_log.search() methods.

- [x] Implementation
- [x] Local Testing
- [x] Unit Tests
- [x] Integration Tests

### DIFF
--- a/pyatlan/client/search_log.py
+++ b/pyatlan/client/search_log.py
@@ -77,7 +77,7 @@ class SearchLogClient:
         sorts: List[SortItem],
     ) -> List[SortItem]:
         """
-        Ensures that sorting by creation timestamp is prioritized for Search Log bulk searches.
+        Ensures that sorting by creation timestamp is prioritized for search Log bulk searches.
         :param sorts: List of existing sorting options.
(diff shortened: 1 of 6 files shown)


Aryamanz29 (approved):
LGTM 🚀 

- I made some changes to your `unit` and `integration` tests to ensure pagination is properly executed within the test cases. Previously, you were only asserting checks like this:

```py
assert len(results.current_page()) == size
assert results.count == total_count
```

This isn't sufficient ^^ as we also need to iterate over the results and confirm that the iteration `count == TOTAL_LOG_ENTRIES` (as returned by the API response) ✅ 

Also, I noticed that several assertions were missing in your unit tests. For example, we should include assertions like:

```py
assert mock_logger.call_count == 0
...
...
assert mock_api_caller._call_api.call_count == 2
```

_**We might need to update the audit unit and integration tests as well to follow the same approach ^^**_

PR #452, opened 2024-12-26
DT-875 Fix issue with private property in constructor

ErnestoLoma:


Aryamanz29 (approved):
Correct, I forgot to update this (it was previously a public field). 🤦 

LGTM, thanks!

PR #454, opened 2024-12-26
 FT-842: Enabled the use of `Retry-After` header value for handling rate-limit retries

Aryamanz29:
https://atlanhq.atlassian.net/browse/FT-821?focusedCommentId=312475


ErnestoLoma (approved):
Wow that was easy. I guess it is a standard.

PR #458, opened 2024-12-30
FT-843: Implement GET operation that takes a list of attributes and relationships

vaibhavatlan:
Breaking Change:
- Changed the default setting of `ignore_relationships = False` to `ignore_relationships = True`

Features: 
- Added `attribute` and `relationships_attributes` params in `AssetClient.get_by_qualified_name()` and  `AssetClient.get_by_guid() ` which uses `fluent search`  and not the default raw api call


pyatlan/client/asset.py
```diff
@@ -366,15 +368,50 @@ def get_by_qualified_name(
...
             "ignoreRelationships": ignore_relationships,
         }
+        if (attributes and len(attributes)) or (
+            relationships_attributes and len(relationships_attributes)
+        ):
+            client = AtlanClient.get_default_client()
+            search = (
+                FluentSearch()
+                .select()
+                .where(CompoundQuery.active_assets())
```
Aryamanz29, line 392: - Can you please verify whether the `GET_ENTITY_BY_UNIQUE_ATTRIBUTE` API endpoint returns only `ACTIVE` assets? If it does, then adding `.where(CompoundQuery.active_assets())` is correct. Otherwise, if it returns both `ACTIVE` and `DELETED` assets, we should avoid adding this as it would result in a different response.

pyatlan/client/asset.py
```diff
@@ -366,15 +368,50 @@ def get_by_qualified_name(
...
+                .where(Asset.QUALIFIED_NAME.eq(qualified_name))
+            )
+            if attributes:
+                for attribute in attributes:
+                    search = search.include_on_results(attribute)
+            if relationships_attributes:
+                for relation_attribute in relationships_attributes:
+                    search = search.include_on_relations(relation_attribute)
```
Aryamanz29, line 400: I believe we can remove the redundant `if` conditions:  
```python
if attributes:
...

if relationships_attributes:
...
```

This is because we already have a check above:  
```python
if (attributes and len(attributes)) or (
        relationships_attributes and len(relationships_attributes)
    ):
    ...
```

pyatlan/client/asset.py
```diff
@@ -406,13 +445,45 @@ def get_by_guid(
...
             "ignoreRelationships": ignore_relationships,
         }
+        if (attributes and len(attributes)) or (
+            relationships_attributes and len(relationships_attributes)
+        ):
+            client = AtlanClient.get_default_client()
+            search = (
+                FluentSearch()
+                .select()
+                .where(CompoundQuery.active_assets())
```
Aryamanz29, line 467: - Can you please verify whether the `GET_ENTITY_BY_GUID ` API endpoint returns only `ACTIVE` assets? If it does, then adding `.where(CompoundQuery.active_assets())` is correct. Otherwise, if it returns both `ACTIVE` and `DELETED` assets, we should avoid adding this as it would result in a different response.

pyatlan/client/asset.py
```diff
@@ -406,13 +445,45 @@ def get_by_guid(
...
+                .where(Asset.GUID.eq(guid))
+            )
+            if attributes:
+                for attribute in attributes:
+                    search = search.include_on_results(attribute)
+            if relationships_attributes:
+                for relation_attribute in relationships_attributes:
+                    search = search.include_on_relations(relation_attribute)
```
Aryamanz29, line 475: I believe we can remove the redundant `if` conditions:  
```python
if attributes:
...

if relationships_attributes:
...
```

This is because we already have a check above:  
```python
if (attributes and len(attributes)) or (
        relationships_attributes and len(relationships_attributes)
    ):
    ...
```

pyatlan/client/asset.py
```diff
@@ -357,7 +357,9 @@ def get_by_qualified_name(
         qualified_name: str,
         asset_type: Type[A],
         min_ext_info: bool = False,
-        ignore_relationships: bool = False,
+        ignore_relationships: bool = True,
+        attributes: Optional[List[str]] = None,
+        relationships_attributes: Optional[List[str]] = None,
```
Aryamanz29, line 362: I think we should update the type to:  

```python
attributes: Optional[Union[List[str], List[AtlanField]]] = None
relationships_attributes: Optional[Union[List[str], List[AtlanField]]] = None
```

This change reflects that the parameters can accept either a string or a search field as input, for example:  

```python
"name"  
Asset.NAME
```

pyatlan/client/asset.py
```diff
@@ -396,7 +433,9 @@ def get_by_guid(
         guid: str,
         asset_type: Type[A] = Asset,  # type: ignore[assignment]
         min_ext_info: bool = False,
-        ignore_relationships: bool = False,
+        ignore_relationships: bool = True,
+        attributes: Optional[List[str]] = None,
+        relationships_attributes: Optional[List[str]] = None,
```
Aryamanz29, line 438: I think we should update the type to:  

```python
attributes: Optional[Union[List[str], List[AtlanField]]] = None
relationships_attributes: Optional[Union[List[str], List[AtlanField]]] = None
```

This change reflects that the parameters can accept either a string or a search field as input, for example:  

```python
"name"  
Asset.NAME
```

Aryamanz29 (changes requested):
Good progress!
Here are a few minor changes:

pyatlan/client/asset.py
vaibhavatlan: No it doesn't. Will remove .where(CompoundQuery.active_assets()).

pyatlan/client/asset.py
vaibhavatlan: No it doesn't. Will remove .where(CompoundQuery.active_assets()).

pyatlan/client/asset.py
vaibhavatlan: Removed

pyatlan/client/asset.py
vaibhavatlan: Removed

pyatlan/client/asset.py
vaibhavatlan: Got it. Changed

pyatlan/client/asset.py
vaibhavatlan: @Aryamanz29 
Encountered an issue when removing the `if` conditions:  
```python
if attributes:
...

if relationships_attributes:
...
```

It raised a `NoneType` error when either `attributes` or `relationships_attributes` was empty. Resolved it using the following approach:  
```python
for attribute in attributes or []:  # type: ignore[union-attr]
...

for relation_attribute in relationships_attributes or []:  # type: ignore[union-attr]
...
```

pyatlan/client/asset.py
vaibhavatlan: Encountered an issue when removing the `if` conditions:  
```python
if attributes:
...

if relationships_attributes:
...
```

It raised a `NoneType` error when either `attributes` or `relationships_attributes` was empty. Resolved it using the following approach:  
```python
for attribute in attributes or []:  # type: ignore[union-attr]
...

for relation_attribute in relationships_attributes or []:  # type: ignore[union-attr]
...
```

Aryamanz29 (approved):
LGTM!

Make sure when writing integration tests to check real-world scenarios (which I later fixed in my latest commits) ✅ 

For example, in your previous integration tests, you only tested for `[attributes=["name", "qualified_name"]` which misses cases for 🧪 

- `related_attributes`
- `attributes` (using `AtlanField`)
- `related_attributes` (using `AtlanField`)

PR #459, opened 2024-12-30
FT-864 Add logic for open telementry logging.

ErnestoLoma:


pyatlan/pkg/utils.py
```diff
@@ -12,6 +12,56 @@
...
+    )
+    from opentelemetry.sdk._logs import (  # type:ignore
+        LogData,
+        LoggerProvider,
+        LoggingHandler,
+    )
+    from opentelemetry.sdk._logs._internal.export import (  # type:ignore
+        BatchLogRecordProcessor,
+    )
+    from opentelemetry.sdk.resources import Resource  # type:ignore
```
Aryamanz29, line 28: Since this feature depends on the external library (`opentelemetry`), could we pin its version in our `requirements.txt`?

pyatlan/pkg/utils.py
```diff
@@ -117,3 +167,54 @@ def validate_connector_and_connection(v):
...
+            endpoint=os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT"), insecure=True
+        )
+        logger_provider.add_log_record_processor(
+            CustomBatchLogRecordProcessor(otel_log_exporter)
+        )
+
+        otel_handler = LoggingHandler(level=level, logger_provider=logger_provider)
+        otel_handler.setLevel(level)
+        logger.addHandler(otel_handler)
+        logger.info("OpenTelemetry handler added to the logger.")
```
Aryamanz29, line 220: Do we need to update `Renderers.pkl` in the `Java SDK` as well (especially to include these new methods in the Python imports), since we’ll be using the `.pkl-based` toolkit to render the custom package? 🤔 

For example: https://github.com/atlanhq/atlan-java/blob/5c116422f231ebb1efa99bcfaf053100f72c2d2c/package-toolkit/config/src/main/resources/Renderers.pkl#L366

pyatlan/pkg/utils.py
Aryamanz29: Got it! 
By default, we don’t provide `opentelemetry` logging support for custom packages (please correct me if I'm wrong)

To enable it, users need to:

1. Install the `opentelemetry` library, which sets `OTEL_IMPORTS_AVAILABLE`.  
2. Import the `add_otel_handler` method and use it like this:  
   ```python
   add_otel_handler(logger, level, resource)
   ```  
   This will attach the new handler. ^^ ✅

ErnestoLoma: No @Aryamanz29. This library should not be in the requirements.txt for the pyatlan. That's why the imports are in a try block. Only the workflows need the libraries. They will be in the requirements.txt that is generated for the workflow along with pyatlan.

pyatlan/pkg/utils.py
ErnestoLoma: Yes I will update those to include the requisite imports in the requirements.txt and the invocation of `add_otel_handler` in the code.

Aryamanz29 (approved):
LGTM :rocket:

PR #464, opened 2025-01-02
FT-679 Modified find_by_type to use regular expression

ErnestoLoma:


Aryamanz29 (approved):
LGTM 🚀 
Tested locally, and it works as expected! ✅

PR #465, opened 2025-01-02
FT-887: Unable to append terms to a particular TableauDatasource

vaibhavatlan:
Updated the `append_terms` method to optimize retrieval by changing the default value of `ignore_relationships` from `False` to `True` in both `get_by_guid()` and `get_by_qualified_name()`. This ensures only minimal information is fetched, as relationship attributes are not required when updating terms.


Aryamanz29 (approved):
LGTM 🚀

Aryamanz29 (changes requested):
Could you please double-check and update the `replace_terms()` and `remove_terms() ` methods as well? It seems these methods also use `get_by_guid()` and `get_by_qualified_name()` with `ignore_relationships = False`? 🙏

Aryamanz29 (changes requested):
**As discussed:**
We can't ignore relationships for these term methods, which results in retrieving all relationships, as the current get_by_guid and get_by_qualified_name methods don't provide a way to retrieve only assigned_terms for an asset. This functionality is now being added in this PR :pullrequest:

**Next steps:**
- Use the current work on the `FT-843` branch by rebasing, or create a new PR for `FT-887`.
- Update these methods to retrieve only the `assigned_terms` of an asset and nothing else.

**_This approach should fix the bug 🐞  while ensuring no breaking changes are introduced to the existing methods 🤞 ✅_**

Aryamanz29 (approved):
LGTM!

- I'm not very inclined towards removing existing unit/integration tests unless there is a deprecation. In our case, we've changed the method implementation using `SaveSemantic`. Instead of [removing existing unit tests,](https://github.com/atlanhq/atlan-python/pull/465/files#diff-8a93cfa9211c8f1cfcf07190aae5a33542431e9906054bcab28e169d71ad133bL293-L372) we should update them according to the current implementation. For now, I'm okay with the removal, but let's open a new ticket to ensure we add those unit tests again to cover the scenarios that were removed ✅

PR #472, opened 2025-01-09
CXD-304 | Support for Dataverse typedefs 

rupali-atlan:
## 🆕 Creator Methods
- Dataverse Asset creator methods
- Addition of Dataverse connector type under SaaS


Aryamanz29 (approved):
LGTM 🚀 
Some changes I made:

- https://github.com/atlanhq/atlan-python/pull/472/commits/7d702f1002794dda7e8e4e5ea6fc0f04f206e929: Updated to use `creator()` and `updater()` instead of the old names (`create` and `create_for_modification`), which have been deprecated since `v2` of `pyatlan`. We should ensure these methods are used when adding new typedefs.
- https://github.com/atlanhq/atlan-python/pull/472/commits/2c09ca54b975319662949adefb1e41acb58ba19c: When adding integrations, we should also include test cases for `update`, `retrieve`, and `deletion` scenarios.

PR #493, opened 2025-01-28
APP-5054: Increased default read timeout to `15` minutes for `AtlanClient`

Aryamanz29:
As discussed, for valid cases in the lineage listing API that may take longer, we've decided to increase the default `read` timeout of the client to 15 minutes.


ErnestoLoma (approved):
LGTM

PR #498, opened 2025-02-06
APP-5297: Make 'test' parameter optional in credential `creator()` method

vaibhavatlan:


pyatlan/client/credential.py
```diff
@@ -37,17 +37,20 @@ def __init__(self, client: ApiCaller):
...
     @validate_arguments
-    def creator(self, credential: Credential) -> CredentialResponse:
+    def creator(
+        self, credential: Credential, test: Optional[bool] = True
+    ) -> CredentialResponse:
         """
         Create a new credential.
 
         :param credential: provide full details of the credential's to be created.
+        :param test: whether to validate the credentials (True) or skip validation (False) before creation.
```
Aryamanz29, line 47:
```suggestion
        :param test: whether to validate the credentials (`True`) or skip validation (`False`) before creation, defaults to `True`.
```

pyatlan/client/credential.py
```diff
@@ -37,17 +37,20 @@ def __init__(self, client: ApiCaller):
         self._client = client
 
     @validate_arguments
-    def creator(self, credential: Credential) -> CredentialResponse:
+    def creator(
+        self, credential: Credential, test: Optional[bool] = True
```
Aryamanz29, line 41: I think we can safely remove `Optional` here ^^ since valid values are `True` or `False`, whereas `Optional` implies `None` is also a possibility 🙏

```suggestion
        self, credential: Credential, test: bool = True
```

pyatlan/client/credential.py
```diff
@@ -37,17 +37,20 @@ def __init__(self, client: ApiCaller):
...
 
         :param credential: provide full details of the credential's to be created.
+        :param test: whether to validate the credentials (True) or skip validation (False) before creation.
         :returns: A CredentialResponse instance.
         :raises ValidationError: If the provided `credential` is invalid.
         """
         raw_json = self._client._call_api(
             api=CREATE_CREDENTIALS.format_path_with_params(),
-            query_params={"testCredential": "true"},
+            query_params={"testCredential": f"{test}"},
```
Aryamanz29, line 53: Can we simply pass a boolean flag (`test`) to this request? I assume the `requests` library will automatically handle query parameters 🤔  Can you please double-check? 🙏  Also, I see we're using `f"{test}"`, which seems incorrect - it converts to string `"True"` or `"False"` instead of `"true"` or  `"false"`. It's still worth checking if we should use a boolean flag (`test`)  here instead of using any strings 🙏 

```suggestion
            query_params={"testCredential": test},
```

Aryamanz29 (approved):
Looks good!
Some doubts below:

pyatlan/client/credential.py
vaibhavatlan: Done

pyatlan/client/credential.py
vaibhavatlan: Done

pyatlan/client/credential.py
vaibhavatlan: Done

PR #503, opened 2025-02-11
APP-5372: Add Validation Checks for Credential Creation in Creator

vaibhavatlan:


pyatlan/client/credential.py
```diff
@@ -47,6 +47,12 @@ def creator(self, credential: Credential, test: bool = True) -> CredentialRespon
         :returns: A CredentialResponse instance.
         :raises ValidationError: If the provided `credential` is invalid.
         """
+
+        if not test and (
+            credential.username is not None or credential.password is not None
+        ):
+            raise ErrorCode.UNABLE_TO_CREATE_CREDENTIAL.exception_with_parameters()
```
Aryamanz29, line 54: How about simplifying this to:

```suggestion
if not test and any((credential.username, credential.password)):
    raise ErrorCode.UNABLE_TO_CREATE_CREDENTIAL.exception_with_parameters()
```

pyatlan/client/credential.py
```diff
@@ -47,6 +47,12 @@ def creator(self, credential: Credential, test: bool = True) -> CredentialRespon
         :returns: A CredentialResponse instance.
         :raises ValidationError: If the provided `credential` is invalid.
         """
```
Aryamanz29, line 49: I would also update the docstrings here to include the this new `InvalidRequestError` ^^

pyatlan/errors.py
```diff
@@ -615,6 +615,13 @@ class ErrorCode(Enum):
         "Check the details of the server's message to correct your request.",
         InvalidRequestError,
     )
+    UNABLE_TO_CREATE_CREDENTIAL = (
+        400,
+        "ATLAN-PYTHON-400-071",
+        "When test is set to False, credentials can only be created without a username and password.",
+        "Either switch test to True or remove the username and password.",
+        InvalidRequestError,
+    )
```
Aryamanz29, line 624: How about simplifying this to:

```suggestion
    UNABLE_TO_CREATE_CREDENTIAL = (
        400,
        "ATLAN-PYTHON-400-071",
        "Credentials cannot include a `username` or `password` when `test` is `False`.",
        "Set `test` to `True` or remove `username` and `password`.",
        InvalidRequestError,
    )
```

Aryamanz29 (changes requested):
Looks good!
Some minor changes below:

pyatlan/client/credential.py
vaibhavatlan: Done

pyatlan/client/credential.py
vaibhavatlan: Done

PR #505, opened 2025-02-12
APP-4750: Add creators for `Quicksight` assets

vaibhavatlan:
- [x]  Jinja Generator Template
- [x] Local testing
- [x] Unit tests
- [x] Integration Tests
- [ ] Docs updated


Aryamanz29 (changes requested):
Looks good!

Some minor changes below:
- Update to use `creator` and `updater` instead of the old names (`create` and `create_for_modification`), which have been deprecated since `v2` of pyatlan. We should ensure these methods are used when adding new typedefs.

- When adding integrations, we should also include test cases for update, retrieve, and deletion scenarios.

Aryamanz29 (approved):
Let's ensure that whenever we copy test cases, we have corresponding tests (`unit` and `integration`) covering the most obvious scenarios 🔎 

For unit tests, we should include `creator()`, `updater()`, and `trim_to_required()`.  
For integration tests, we should cover the full CRUD flow ✅ 

I noticed that the current integration tests don’t include `delete` and `restore` scenarios like other [asset tests](https://github.com/atlanhq/atlan-python/blob/d525f65a2db928486e11e49c185b939446a0cb1d/tests/integration/kafka_asset_test.py#L170-L220). While this is fine for now, let's maintain consistency across all new assets by including these cases whenever we add new ones!

vaibhavatlan:
Got it! I might have missed it earlier. I'll create a Jira ticket for this and include the details above. I'll make sure to keep this in mind when working on asset creation next time. Thanks! 🙏

PR #530, opened 2025-02-17
APP-5338 | Add connector types CrateDB and KX

pranay-at-atlan:
Introduces 2 new connector types and their corresponding icons

[Frontend PR](https://github.com/atlanhq/atlan-frontend/pull/23806)

### DIFF
--- a/pyatlan/model/enums.py
+++ b/pyatlan/model/enums.py
@@ -347,7 +347,8 @@ class AtlanConnectorType(str, Enum):
     SHARED_DRIVE = ("shared-drive", AtlanConnectionCategory.OBJECT_STORE)
     SHARE_POINT = ("share-point", AtlanConnectionCategory.SAAS)
     RDS = ("rds", AtlanConnectionCategory.WAREHOUSE)
-
+    CRATEDB = ('cratedb', AtlanConnectionCategory.DATABASE)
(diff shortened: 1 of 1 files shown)


Aryamanz29 (approved):
LGTM 🚀

pyatlan/model/enums.py
```diff
@@ -1890,6 +1892,8 @@ class AtlanIcon(str, Enum):
     YARN = "PhYarn"
     YIN_YANG = "PhYinYang"
     YOUTUBE_LOGO = "PhYoutubeLogo"
+    CRATEDB = "CrateDb"
+    KX = "Kx"
```
Aryamanz29, line 1896: Are you sure we need to update this? 🤔 
I think we only need to add new connector entries to the `AtlanConnectorType` enum, which you've already done above^^ 

For example
- https://github.com/atlanhq/atlan-python/pull/502/files
- https://github.com/atlanhq/atlan-frontend/pull/23719/files

Aryamanz29 (changes requested):
One doubt here:

pyatlan/model/enums.py
Aryamanz29: I've removed this for now since I need to cut the SDK release. If we need this, please create a PR in the future. Thanks!

PR #535, opened 2025-02-19
APP-5412 : Change the Query Model to use Pydantic model

vaibhavatlan:


pyatlan/model/search.py
```diff
@@ -1893,8 +1905,21 @@ def validate_query(cls, v, values):
         else:
             raise ValueError("Either query or post_filter is required")
 
+    # Transforming dictionaries into SortItem instances to facilitate deserialization.
+    # This conversion occurs prior to invoking the initializer and validator_sort(),
+    # preventing potential validation-related issues from arising.
```
Aryamanz29, line 1910: I would simplify it and include it as a method docstring:

```suggestion
"""
Convert dictionary entries into SortItem instances before validation  
to ensure proper deserialization and prevent validation issues.
"""
```

pyatlan/model/search.py
```diff
@@ -1893,8 +1905,21 @@ def validate_query(cls, v, values):
...
+    @validator("sort", pre=True)
+    def dict_to_SortItem(cls, sort):
```
Aryamanz29, line 1912: Since we're using `SortItem.from_dict`, I would rename the validator to `sortitem_from_dict()` for clarity:

```suggestion
    def sortitem_from_dict(cls, sort):
```

pyatlan/model/search.py
```diff
@@ -1893,8 +1905,21 @@ def validate_query(cls, v, values):
...
+        if all(isinstance(item, dict) for item in sort):
+            sort = [SortItem.from_dict(item) for item in sort]
+        return sort
+
     @validator("sort", always=True)
     def validate_sort(cls, sort, values):
+
+        if all(isinstance(item, dict) for item in sort):
+            sort = [SortItem.from_dict(item) for item in sort]
```
Aryamanz29, line 1921: Do we still need these even though we already handled this in the `sortitem_from_dict()` validator above? ^^ 🤔

tests/unit/test_search_model.py
```diff
@@ -293,6 +293,17 @@ def test_index_search_request():
...
+    # deserialization test
+    data = {
+        "query": {"term": {"__typeName.keyword": {"value": "Schema"}}},
+        "from": 0,
+        "size": 300,
+        "track_total_hits": True,
+        "sort": [],
+    }
+    test_dsl = DSL(**data)
+    assert test_dsl.query == data["query"]  # TODO: <-- make sure its Query object
```
Aryamanz29, line 305:
```py
class DSL(AtlanObject):
    from_: int = Field(default=0, alias="from")
    size: int = Field(default=300)
    aggregations: Dict[str, Aggregation] = Field(default_factory=dict)
    track_total_hits: Optional[bool] = Field(default=True, alias="track_total_hits")
    post_filter: Optional[Query] = Field(default=None, alias="post_filter")
    query: Optional[Query]
    req_class_name: Optional[str] = Field(default=None, exclude=True)
    sort: List[SortItem] = Field(default_factory=list, alias="sort")
```

This sample test was written to test the logic; it's not complete. Please move it to a separate unit test case and ensure it includes a complex (nested) data dictionary that covers all DSL model fields ^^ 🙏 
 
```suggestion
    data = {
        "query": {"term": {"__typeName.keyword": {"value": "Schema"}}},
        "from": 0,
        "size": 300,
        "track_total_hits": True,
        "sort": [],
    }
    test_dsl = DSL(**data)
    assert test_dsl.query == data["query"] 
```

Aryamanz29: Okay, since we already have `test_dsl_serialization_and_deserialization`, please remove the above test. ^^

tests/unit/test_search_model.py
```diff
@@ -1558,3 +1569,29 @@ def test_with_active_term():
...
+def test_dsl_serialization_and_deserialization():
+
+    dsl1 = DSL(
+        from_=0,
+        aggregations={"test1": {"test2": {"field": "__test_field"}}},
+        size=300,
+        sort=[SortItem(field="created", order=SortOrder.ASCENDING, nested_path="test")],
+        query=Term(field="__typeName.keyword", value="Schema"),
+        track_total_hits=False,
+    )
```
Aryamanz29, line 1583: Let's also ensure we test this on complex data structures in unit tests, e.g., nested `aggregations`, multiple `sort` items in a list, and multiple queries of different types (`term`, `bool`, `range`, etc).

```suggestion
   dsl_through_model = DSL(
        from_=0,
        aggregations={"test1": {"test2": {"field": "__test_field"}}},
        size=300,
        sort=[SortItem(field="created", order=SortOrder.ASCENDING, nested_path="test")],
        query=Term(field="__typeName.keyword", value="Schema"),
        track_total_hits=False,
    )
```

tests/unit/test_search_model.py
```diff
@@ -1558,3 +1569,29 @@ def test_with_active_term():
...
+
+    data = {
+        "aggregations": {"test1": {"test2": {"field": "__test_field"}}},
+        "query": {"term": {"__typeName.keyword": {"value": "Schema"}}},
+        "from": 0,
+        "size": 300,
+        "track_total_hits": False,
+        "sort": [{"created": {"order": "asc", "nested": {"path": "test"}}}],
+    }
+    test_dsl = DSL(**data)
```
Aryamanz29, line 1593: Let's also ensure we test this on complex data structures in unit tests, e.g., nested `aggregations`, multiple `sort` items in a list, and multiple queries of different types (`term`, `bool`, `range`, etc).

```suggestion
   raw_dsl_data = {
        "aggregations": {"test1": {"test2": {"field": "__test_field"}}},
        "query": {"term": {"__typeName.keyword": {"value": "Schema"}}},
        "from": 0,
        "size": 300,
        "track_total_hits": False,
        "sort": [{"created": {"order": "asc", "nested": {"path": "test"}}}],
    }
   dsl_through_raw = DSL(**raw_dsl_data)
```

tests/unit/test_search_model.py
```diff
@@ -1558,3 +1569,29 @@ def test_with_active_term():
...
+
+    assert test_dsl.json(exclude_unset=True, by_alias=True) == dsl1.json(
+        exclude_unset=True, by_alias=True
+    )
```
Aryamanz29, line 1597: We should also assert other serialization methods and params, such as:
- `.dict()`
- `.dict(exclude_unset=True, by_alias=True))`
- `.json()`

Aryamanz29 (changes requested):
Looks good!
Some minor req changes below:

- Let's also rebase (ie: [squash commits ](https://www.datacamp.com/tutorial/git-squash-commits) these experimental commits (we don't want them in our commit history), for example:

- [testing](https://github.com/atlanhq/atlan-python/pull/535/commits/449f13c7a0193d58045bc20705013c0197af45a8), [[temp] some experiments](https://github.com/atlanhq/atlan-python/pull/535/commits/99c10b56b507c881ff4c4764194525445dd2b3cb) etc...

pyatlan/model/search.py
vaibhavatlan: Done

pyatlan/model/search.py
vaibhavatlan: Done

tests/unit/test_search_model.py
vaibhavatlan: Got it

tests/unit/test_search_model.py
vaibhavatlan: Done

tests/unit/test_search_model.py
vaibhavatlan: Done

tests/unit/test_search_model.py
vaibhavatlan: Done

Aryamanz29 (approved):

PR #539, opened 2025-02-21
APP-4915 : Provide helper to translate DataProduct's DSL to an asset listing

vaibhavatlan:


tests/integration/data_mesh_test.py
```diff
@@ -486,6 +487,19 @@ def test_find_product_by_name(client: AtlanClient, product: DataProduct):
...
+def test_product_get_assets(client: AtlanClient, product: DataProduct):
+    test_product = client.asset.get_by_guid(
+        product.guid, asset_type=DataProduct, ignore_relationships=False
+    )
+    assert test_product
+    assert test_product.data_product_assets_d_s_l
+    asset_list = test_product.get_assets()
+
+    assert asset_list.count > 0
+    assert isinstance(asset_list, IndexSearchResults)
```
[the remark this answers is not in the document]
Aryamanz29: Something like: https://github.com/atlanhq/atlan-python/blob/9ec885ee55f80e48960654c9f38ece16eb9f580b/tests/integration/test_index_search.py#L338-L351

Aryamanz29 (changes requested):
Looks good!
Some minor improvements:

pyatlan/model/assets/core/data_product.py
```diff
@@ -100,6 +101,16 @@ def updater(
             )
         return product
 
+    def get_assets(self):
+        from pyatlan.client.atlan import AtlanClient
+
+        client = AtlanClient().get_default_client()
+        dp_dsl = self.data_product_assets_d_s_l
+        json_object = json.loads(dp_dsl)
+        request = IndexSearchRequest(**json_object["query"])
```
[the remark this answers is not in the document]
vaibhavatlan: Done

pyatlan/model/assets/core/data_product.py
```diff
@@ -100,6 +101,24 @@ def updater(
             )
         return product
 
+    def get_assets(self, client: Optional['AtlanClient'] = None):
```
Aryamanz29, line 104: Why are we using `'AtlanClient'` as a string here? We need the actual type. If it's causing a circular `ImportError`, we should handle it like this:

```py
from typing import TYPE_CHECKING
if TYPE_CHECKING:
 ```

```suggestion
    def get_assets(self, client: Optional[AtlanClient] = None):
```

pyatlan/model/assets/core/data_product.py
```diff
@@ -100,6 +101,24 @@ def updater(
...
+        """
+        Reterieves list of all assets linked to the provided data product.
+
+        :param client: connectivity to an Atlan tenant
+
+        :raises AtlanError: if there is an issue interacting with the API
+        :returns: Index Search Results with list of all assets linked to the provided data product
```
Aryamanz29, line 111:
```suggestion
        :returns: instance of `IndexSearchResults` with list of all assets linked to the provided data product
```

pyatlan/model/assets/core/data_product.py
```diff
@@ -100,6 +101,24 @@ def updater(
             )
         return product
 
+    def get_assets(self, client: Optional['AtlanClient'] = None):
+        """
+        Reterieves list of all assets linked to the provided data product.
+
+        :param client: connectivity to an Atlan tenant
```
Aryamanz29, line 108:
```suggestion
        :param client: connectivity to an Atlan tenant (optional). If not provided, the default client will be used.
```

pyatlan/model/assets/core/data_product.py
```diff
@@ -100,6 +101,24 @@ def updater(
...
+
+        :raises AtlanError: if there is an issue interacting with the API
+        :returns: Index Search Results with list of all assets linked to the provided data product
+
+        """
+        from pyatlan.client.atlan import AtlanClient
+        client = AtlanClient.get_default_client() if not client else client
+        dp_dsl = self.data_product_assets_d_s_l
+        json_object = json.loads(dp_dsl) if dp_dsl else {}
```
Aryamanz29, line 117:
```suggestion
        dp_dsl = self.data_product_assets_d_s_l or {}
        json_object = json.loads(dp_dsl)
```

tests/integration/data_mesh_test.py
```diff
@@ -486,6 +487,23 @@ def test_find_product_by_name(client: AtlanClient, product: DataProduct):
...
 
+@pytest.mark.order(after="test_retrieve_product")
+def test_product_get_assets(client: AtlanClient, product: DataProduct):
+    test_product = client.asset.get_by_guid(
+        product.guid, asset_type=DataProduct, ignore_relationships=False
+    )
+    assert test_product
+    assert test_product.data_product_assets_d_s_l
+    asset_list = test_product.get_assets()
+    TOTAL_ASSETS = asset_list.count
```
Aryamanz29, line 498:
```suggestion
    assert asset_list.count and asset_list.count > 0
    TOTAL_ASSETS = asset_list.count
```

Aryamanz29 (changes requested):
Looks good, some minor changes below:

pyatlan/model/assets/core/data_product.py
vaibhavatlan: @Aryamanz29 
The reason I wrote this 

>         json_object = json.loads(dp_dsl) if dp_dsl else {}

Was because mypy test raises this issue:


> pyatlan/model/assets/core/data_product.py:118: error: Argument 1 to "loads" has incompatible type "str | None"; expected "str | bytes | bytearray"  [arg-type]
> Found 1 error in 1 file (checked 599 source files)

So doing what you suggested above won't resolve this issue.

PR #545, opened 2025-02-25
[gen/templates] Fixed generator templates (as per recent updates)

Aryamanz29:


Aryamanz29:
@vaibhavatlan / @rupali-atlan Please ensure that whenever we make any changes (e.g: adding `creator`, `updater`, etc.) to `pyatlan/model/assets/*`, we also update the corresponding generator templates (`.jinja2`) 🙏  
Otherwise, when the generator is re-run to extend the SDK with new typedef models, your changes will be overwritten and lost ❌ 

Luckily, I caught these missing template updates ^^ through integration tests -- otherwise, this could have introduced unintended breaking changes, which we should always avoid 🤞

PR #548, opened 2025-02-27
APP-5657 : Add support for `MatchPhase` query to the SDKs

vaibhavatlan:


Aryamanz29 (approved):
LGTM 🚀

Aryamanz29 (commented):
Can we update the dev-portal to include how to use the new `MatchPhrase`?  
[Developer Docs - Search Queries](https://developer.atlan.com/search/queries/text)

Aryamanz29 (changes requested):
Looks good!  

Can you please add a test case for the new `TextField.match_phrase()` method as well? (Maybe using `FluentSearch()`)? 🙏

Aryamanz29 (approved):

PR #569, opened 2025-03-13
APP-5740 : Add support for iterative pagination to `User`, `Group`, and `Workflow` results

vaibhavatlan:


pyatlan/client/group.py
```diff
@@ -152,9 +152,10 @@ def get_all(
...
-        if response := self.get(offset=offset, limit=limit, sort=sort, columns=columns):
-            return response.records  # type: ignore
-        return None  # type: ignore
+        response: GroupResponse = self.get(
+            offset=offset, limit=limit, sort=sort, columns=columns
+        )
+        return response  # type: ignore
 
     @validate_arguments
     def get_by_name(
```
Aryamanz29, line 161: Let's update this method as well, since it also returns a list of `AtlanGroup`. We should have it return a `GroupResponse` instead, which includes support for iterative pagination.

pyatlan/client/user.py
```diff
@@ -211,7 +211,7 @@ def get_all(
         :returns: a list of all the users in Atlan
         """
         response: UserResponse = self.get(offset=offset, limit=limit, sort=sort)
-        return [user for user in response]
+        return response
 
     @validate_arguments
     def get_by_email(
```
Aryamanz29, line 217: Let's update this method as well, since it also returns a list of `AtlanUser`. We should have it return a `UserResponse` instead, which includes support for iterative pagination.

Aryamanz29: Same for `get_by_emails()` method 👇

Aryamanz29 (changes requested):
Good progress!

It looks like I had already added iterative pagination to `user` and `group` in the past, but it seems that proper testing of those methods is missing. For example, `GroupClient.get_members` is correctly implemented and returns a `UserResponse`, but `GroupClient.get_all()` just returns `response.records`, which doesn't support pagination since it's embedded in the response classes (e.g: `GroupResponse`) 🤔 ✅ 

-------

I’ve tested the changes manually - all looks good! (few requested changes below)

#### Requested Changes

* **Add integration tests** for the remaining `user` and `group` methods where pagination has been implemented.
  *(We can skip `workflow` for now since it may require re-running workflows multiple times, which could add 1–2 minutes to the integration test job - but make sure it's manually tested locally.)*

* **Pagination test strategy**:

  * Use a **small page size** to ensure multiple API calls are triggered.
  * To determine the small page size dynamically:

    * First, make an API call with `size: 0` to get the **total record count**.
    * Then, calculate page size as:

      ```python
      size = max(1, math.ceil(TOTAL_ASSETS / 5))
      ```

* **Update documentation** for all these pagination methods to reflect their usage and any specifics.

`Example`:

https://github.com/atlanhq/atlan-python/blob/7340bbf75e0c2d9713f77a8843e37571560a4428/tests/integration/test_index_search.py#L351-L365

pyatlan/client/user.py
```diff
@@ -291,11 +291,10 @@ def get_by_usernames(
...
         """
         username_filter = '{"username":{"$in":' + dumps(usernames or [""]) + "}}"
-        if response := self.get(
+        response: UserResponse = self.get(
             offset=offset, limit=limit, post_filter=username_filter
-        ):
-            return response.records
-        return None
+        )
+        return response.records  # type: ignore
```
Aryamanz29, line 297: Shouldn't it just be response? 🤔

vaibhavatlan: yes. Just noticed it. Fixed it in the last commit.

Aryamanz29 (approved):

PR #573, opened 2025-03-17
APP-5519: Migrated `AtlanClient._default_client` to thread-local storage + caches now bound to optional client

Aryamanz29:
This PR fixes `pyatlan` caching issues in concurrent environments:
https://atlanhq.atlassian.net/browse/APP-5519?focusedCommentId=367914

References for why we opted for `TLS` to solve this:
https://docs.python.org/3/library/threading.html
https://superfastpython.com/thread-local-data
https://stackoverflow.com/a/105025
https://stackoverflow.com/questions/1408171/thread-local-storage-in-python

### DIFF
--- a/pyatlan/cache/abstract_asset_cache.py
+++ b/pyatlan/cache/abstract_asset_cache.py
@@ -4,11 +4,14 @@ from __future__ import annotations
 
 import threading
 from abc import ABC, abstractmethod
-from typing import Any
+from typing import TYPE_CHECKING, Any, Dict
 
 from pyatlan.errors import ErrorCode
 from pyatlan.model.assets import Asset
 from pyatlan.model.enums import AtlanConnectorType
+
+if TYPE_CHECKING:
+    from pyatlan.client.atlan import AtlanClient
 
 
 class AbstractAssetCache(ABC):
@@ -17,17 +20,12 @@ class AbstractAssetCache(ABC):
     to all caches, where a cache is populated entry-by-entry.
     """
 
-    def __init__(self, client):
+    def __init__(self, client: AtlanClient):
         self.client = client
         self.lock = threading.Lock()
-        self.name_to_guid = dict()
-        self.guid_to_asset = dict()
-        self.qualified_name_to_guid = dict()
-
-    @classmethod
-    @abstractmethod
-    def get_cache(cls):
-        """Abstract method to retreive cache."""
+        self.name_to_guid: Dict[str, str] = dict()
+        self.guid_to_asset: Dict[str, Asset] = dict()
+        self.qualified_name_to_guid: Dict[str, str] = dict()
 
     @abstractmethod
     def lookup_by_guid(self, guid: str):
@@ -84,9 +82,9 @@ class AbstractAssetCache(ABC):
         name = asset and self.get_name(asset)
         if not all([name, asset.guid, asset.qualified_name]):
             return
-        self.name_to_guid[name] = asset.guid
-        self.guid_to_asset[asset.guid] = asset
-        self.qualified_name_to_guid[asset.qualified_name] = asset.guid
+        self.name_to_guid[name] = asset.guid  # type: ignore[index]
+        self.guid_to_asset[asset.guid] = asset  # type: ignore[index]
+        self.qualified_name_to_guid[asset.qualified_name] = asset.guid  # type: ignore[index]
 
     def _get_by_guid(self, guid: str, allow_refresh: bool = True):
         """
--- a/pyatlan/cache/atlan_tag_cache.py
+++ b/pyatlan/cache/atlan_tag_cache.py
@@ -2,7 +2,7 @@
 # Copyright 2025 Atlan Pte. Ltd.
 from __future__ import annotations
 
-from threading import Lock, local
+from threading import Lock
 from typing import TYPE_CHECKING, Dict, Optional, Set
 
 from pyatlan.errors import ErrorCode
@@ -13,7 +13,6 @@ if TYPE_CHECKING:
     from pyatlan.client.atlan import AtlanClient
 
 lock: Lock = Lock()
-atlan_tag_cache_tls = local()  # Thread-local storage (TLS)
 
 
 class AtlanTagCache:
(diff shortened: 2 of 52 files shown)


Aryamanz29:
> I'm a bit confused. I thought we were getting rid of the get_default_client method on the AtlanClient class.
Also since the caches have a reference to the the client and the client has a reference to the caches, I don't know why we need the get_cache or refresh_cache class methods Seems to me that the client should just lazily create an instance of the cache as needed. Am I missing something?

Yes @ErnestoLoma - in the last standup, we all agreed to completely remove `get_default_client`. Since it's a `ClassVar`, it causes issues in concurrent/multi-threaded environments, as it's shared across threads, affecting its value and cache logic, leading to inconsistent behavior reported by users.  

I tried removing all instances of `get_default_client` and the `cache_key` logic in `get_cache` methods. However, I got stuck in `AtlanTagName` due to Pydantic pre-validators.  

For example, the `Purpose` asset has an `AtlanTagName` field, and its pre-validator converts tag hash strings into human-readable display text. Since pre-validators run before instance creation, they call `AtlanTagCache` to resolve tag hashes, requiring a client to make API calls. But the client is an instance variable (`AtlanTagName(client)`), making it inaccessible to class methods and pre-validators.  

And if we make the client as a `ClassVar` to fix this ^^ then would be equivalent to using `get_default_client()`, which defeats the purpose of per-cache client management. Instead, I propose making the client optional in the constructor while ensuring the default client remains thread-safe using TLS. This approach maintains SDK backward compatibility ✅

-----

Most of the cache inconsistency issues reported by users (and which I was able to replicate) stem from `_default_client` being a `ClassVar`, making client switches non-thread-safe.  

You might wonder why there are so many client switches now <-- this is due to the recent `JWT` token timeout change in production. We now expire API tokens after `15` minutes, and the SDK's token refresh mechanism creates a new client after expiration, leading to a new cache key.  

Since `caches (dict)` is also a `ClassVar` in different cache classes, this further compounds the issue. However, now that we've migrated both `_default_client` and `caches (dict)` from `ClassVar`  to `TLS`, we're safe to use our `cache_key` logic again, which also helps reduce redundant API calls to populate caches.  

I've also tested these changes across multiple scenarios (switching clients, multiple clients isolated in threads), and everything works as expected without introducing any breaking changes. Hopefully, this will fix most issues, but we’ll see what new challenges arise after releasing this change 💪  - optimistic for positive results!  🤞 🚀

-----

```py
  @classmethod
    def __get_validators__(cls):
        yield cls._convert_to_display_text
   
   @classmethod
    def _convert_to_display_text(cls, data):
        from pyatlan.cache.atlan_tag_cache import AtlanTagCache

        if isinstance(data, AtlanTagName):
            return data
        if display_text := AtlanTagCache.get_name_for_id(data):
            return AtlanTagName(display_text)
        else:
            return cls.get_deleted_sentinel()
 ```

```py
      purpose_asset = {
        "typeName": "Purpose",
        "attributes": {
            # AtlanTagName
            "purposeClassifications": [
                "some-hash-string1",
                "some-hash-string1",
            ],
        },
    # Verify that deleted tags are correctly set to `None`
    # assert purpose.atlan_tags == [AtlanTagName('(DELETED)')]
    assert purpose.atlan_tags and len(purpose.atlan_tags) == 2
    assert purpose.atlan_tags[0].type_name.__repr__() == f"AtlanTagName('{DELETED_}')"
    assert purpose.atlan_tags[1].type_name.__repr__() == f"AtlanTagName('{DELETED_}')"
    assert purpose.purpose_atlan_tags and len(purpose.purpose_atlan_tags) == 2
    assert purpose.purpose_atlan_tags[0].__repr__() == f"AtlanTagName('{DELETED_}')"
    assert purpose.purpose_atlan_tags[1].__repr__() == f"AtlanTagName('{DELETED_}')"
```

ErnestoLoma:
> I tried removing all instances of `get_default_client` and the `cache_key` logic in `get_cache` methods. However, I got stuck in `AtlanTagName` due to Pydantic pre-validators.

@Aryamanz29 so if the problem is limited to the _convert_to_display_text function then the solution should focus there and not in the AtlanTagCache. If we get  `_default_client` in that block then we can retrieve the cache from the client. Plus there's a fundamental problem in the way we are setting the setting `_default_client` as the following block of code demonstrates:
```
import os

from pyatlan.model.assets import (
    Table,
)

client = AtlanClient(base_url="https://fieldtec41.atlan.com")
client2 = AtlanClient(base_url="https://field-sandbox.atlan.com", api_key=os.environ["OTHER_API_KEY"])
table = client.asset.get_by_guid(guid="47835445-20c4-438f-af23-c4f0f8f491e2", asset_type=Table)
print(table.atlan_tags)

```
If we run this code it fails because we are retrieving the asset from fieldtec41 but when the attempt is made to get the display text for the tag the _default_client is set to field-sandbox so a check for the tag is made in field sandbox and the tag doesn't exist there. If the line starting with client2 is commented out the code runs fine. 
Currently we are setting `_default_client` when a client is created, instead what we should be setting `_default_client` to `self` whenever a method of `AtlanClient` is invoked. 

The validator could then be changed to the following:
```
 @classmethod
  def _convert_to_display_text(cls, data):

        if isinstance(data, AtlanTagName):
            return data
        if display_text := AtlanClient.get_default_client().atlan_tag_cache.get_name_for_id(data):
            return AtlanTagName(display_text)
        else:
            return cls.get_deleted_sentinel()
```
The name of the variable should also be changed from _default_client to _current_client to better reflect what it is. 
I believe this would eliminate the need for the class methods on the various caches as well as a need for the cache_key since each client has it's own set of caches.
 Also why do we create a new client when the token is refreshed instead of just updating the associated `Session` header?
cc: @cmgrote

ErnestoLoma (changes requested):
See comments in conversation.

Aryamanz29:
Thanks @ErnestoLoma for the review!  🙌 

1. Removed `cache_key` maintenance from caches since it's no longer needed. All public cache methods are now instance methods instead of `@classmethod`.  

2. Updated unit and integration tests accordingly and fixed mypy errors. Some tests (`source_tag_cache`, `connection_cache`, redundant mocks) still need refactoring, which I'll address tomorrow — please ignore them for now.

3. Now, `_current_client` is set whenever a user invokes `@property` methods on `AtlanClient` (e.g:`client.user.*`, `client.group.*`, `client.atlan_tag_cache.*`). I've tested your code snippet, and it works as expected ☝️ ✅   

#### Regarding your point about `_default_client`, while we no longer set it in `AtlanClient.__init__`:

>Currently we are setting _default_client when a client is created, instead what we should be setting _default_client to self whenever a method of AtlanClient is invoked.

There is a necessity for cases where no `@property` methods are called e.g:`client.user.*`, `client.group.*`, `client.atlan_tag_cache.*`) - e.g: resolving classification names via `AtlanTagCache`. Without this, `ErrorCode.NO_ATLAN_CLIENT_AVAILABLE` would be raised 👇

_Therefore, we might still need to [set it in the `AtlanClient` constructor ](https://github.com/atlanhq/atlan-python/pull/573/files#diff-500c50175051b3516ef627be78118b1896998f2fb5cdac266dd028c5e58f2c28R225)**+** [in client `@property` methods](https://github.com/atlanhq/atlan-python/pull/573/files#diff-500c50175051b3516ef627be78118b1896998f2fb5cdac266dd028c5e58f2c28R234)  (which, in any case, sets it again and resolves the error in your provided code snippet above ⬆️ )._

---

```py
from pyatlan.model.assets import Table, Referenceable

client = AtlanClient()

referenceable = Referenceable()
referenceable.classification_names = [123, "456"]

print(referenceable.atlan_tag_names)
```

```py
class Referenceable
...
...
   @property
    def atlan_tag_names(self) -> List[str]:
        from pyatlan.model.constants import DELETED_

       # We need current client here -> AtlanClient.get_current_client() to call `AtlanTagCache`
        if self.classification_names:
            return [
                AtlanClient.get_current_client().atlan_tag_cache.get_name_for_id(tag_id)
                or DELETED_
                for tag_id in self.classification_names
            ]
        return []
 ```

PR #583, opened 2025-03-27
APP-6125 : Pyatlan changes to token refresh

vaibhavatlan:


Aryamanz29 (approved):
We've introduced this change to support `401` token handling in a multithreaded environment, which is used by `longtail` packages.

PR #592, opened 2025-04-07
APP-6218: Improve 401 token refresh handling

Aryamanz29:
- Add delay after token refresh to avoid empty typedef responses.
- Reset `has_retried` if last retry wasn't a `401` to allow future retries.


prateekrai-atlan (approved):
LGTM

PR #595, opened 2025-04-08
APP-5915 : Add "Workflow Name" as an optional param in credentials.get_all()

vaibhavatlan:
For all workflows provided by **Atlan**, the naming convention follows this pattern:

- **Workflow Name:**  
  Example:  
  ```
  atlan-bigquery-1735837155
  ```

- **Credential Name Pattern:**  
  ```
  default-<connector-name>-<timestamp>-0
  ```  
  So for the example above, the credential name would be:  
  ```
  default-bigquery-1735837155-0
  ```

---

For **CSA packages** using credentials, the format is slightly different:

- **Credential Name Pattern:**  
  ```
  <workflow-name>-0
  ```
  
  Did local testing as well  ✅  : 

Testing with Atlan connecter:  
image
image

Cross check the creds here: https://field-sandbox.atlan.com/workflows/profile/atlan-bigquery-1735837155/config

Testing with csa connector:
image
image

Cross check the creds here : https://field-sandbox.atlan.com/workflows/profile/csa-s3-1734926030/config

### DIFF
--- a/pyatlan/client/credential.py
+++ b/pyatlan/client/credential.py
@@ -83,6 +83,7 @@ class CredentialClient:
         filter: Optional[Dict[str, Any]] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
+        workflow_name: Optional[str] = None,
     ) -> CredentialListResponse:
         """
         Retrieves all credentials.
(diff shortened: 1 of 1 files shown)


Aryamanz29 (approved):
LGTM 🚀

PR #598, opened 2025-04-10
APP-5964 : Overload `client.workflow.monitor()` to directly accept `workflow ID/name`

vaibhavatlan:


Aryamanz29 (approved):
LGTM (don’t forget to update the docs for this change as well! 📚)

- I’ve refactored the method to simplify how `workflow_name` is extracted.
- Also added an early return when `name` is not available, along with a `log.info` message to indicate the skip - makes the flow cleaner and easier to follow.
https://refactoring.guru/replace-nested-conditional-with-guard-clauses

PR #600, opened 2025-04-14
APP-6310: Added support for `purpose` and `persona` in `IndexSearchRequest`

Aryamanz29:


hrushikeshdokala (approved):
thanks! its working with the usecase.

PR #602, opened 2025-04-17
fix the indentation issue in documentDB jinja template

pranay-at-atlan:
Fixed the new line issue in template because of which the generator python script was giving indentation error


Aryamanz29 (approved):
LGTM

PR #922, opened 2026-05-01
fix: DbtProcess/DbtColumnProcess parents + typing.get_type_hints shim (SHA-887)

Aryamanz29:
Three SHA-887 follow-ups that all independently break atlan-publish-app's ordering graph against pyatlan v9.x. Bundling them because the symptoms overlap (publish-app order regressions reproduce until the full set lands) and they share the same generator-level mechanism.

## 1. ``DbtProcess`` extends ``Process``, not ``Dbt``

- Atlas typedef: ``superTypes=[\"Dbt\", \"Process\"]``
- Generator picks ``super_types[0]`` → ``Dbt`` → publish-app's ``issubclass(cls, Process)`` check (used to identify lineage types that must publish *after* the entities they reference) returned ``False``.
- ``DbtProcess`` then landed in Batch 1 alongside its upstream entities → ATLAS-404 on ``inputs`` / ``outputs`` lookups in live dbt runs.
- Fix: add ``\"DbtProcess\": \"Process\"`` to ``_SUPERCLASS_OVERRIDES`` in ``class_generator.py`` and patch the generated ``pyatlan/model/assets/dbt_process.py`` source.

## 2. ``DbtColumnProcess`` extends ``ColumnProcess``, not ``Dbt``

- Same root cause: ``superTypes=[\"Dbt\", \"ColumnProcess\"]``.
- Same fix applied to ``pyatlan/model/assets/dbt_column_process.py`` and the override map.

## 3. ``typing.get_type_hints()`` shim for core ``Attributes`` classes

- Five core asset modules (``s_q_l``, ``schema``, ``data_product``, ``data_quality_rule``, ``process``) gate cross-module relationship-type imports under ``if TYPE_CHECKING:`` to avoid circular imports at module load. With ``from __future__ import annotations``, those forward-ref strings can't be resolved by ``typing.get_type_hints()`` against the defining module's ``__dict__`` and the call raises ``NameError`` — e.g. ``name 'DbtTest' is not defined`` for ``Table.Attributes``.
- Pydantic v1's ``update_forward_refs(**localns)`` in ``core/__init__.py`` already handles this for the SDK's own use, but reflection-based downstream tooling (notably ``atlan-publish-app``'s ordering graph) falls through to a degraded fallback that collapses all SQL types into one publish batch.
- Fix: after every asset has loaded, ``core/__init__.py`` injects the resolved classes back into each affected module's ``__dict__`` so ``typing.get_type_hints(cls.Attributes)`` succeeds without a caller-supplied ``localns``.
- A new ``_RUNTIME_INJECTION_OVERRIDES`` constant + ``_build_runtime_injection_block`` helper in ``class_generator.py`` plus a corresponding placeholder in ``templates/core/init.jinja2`` ensure the shim survives regeneration.

## Tests

- ``tests/unit/model/dbt_process_test.py`` — 7 cases (``issubclass`` guards, ``__bases__[0]``, ``type_name`` immutability, ``dbt_*`` + ``Process`` attribute round-trip).
- ``tests/unit/model/dbt_column_process_test.py`` — same 7 cases for the ``ColumnProcess`` parent.
- ``tests/unit/model/get_type_hints_compat_test.py`` — parametrised over all 5 affected core classes plus a concrete spot-check that ``SQL.Attributes.dbt_tests`` resolves to the real ``DbtTest`` class.

## Verified

- [x] ``./qa-checks`` — ruff format / ruff check / mypy all pass
- [x] ``pytest tests/unit -q --ignore=tests/unit/aio`` — 6033 passed, 2 skipped
- [x] Manual ``issubclass`` / ``get_type_hints`` checks against installed pyatlan
- [ ] Downstream verification in atlan-publish-app once this ships: ``DbtProcess`` / ``DbtColumnProcess`` move to a strictly later batch than the entities they reference, and ``typing.get_type_hints()`` resolves on every core ``Attributes`` class without a workaround ``localns``

## Notes

Release notes will be added in the next version bump — ``HISTORY.md`` is left alone here so the already-shipped 9.7.0 section isn't retroactively edited.

### DIFF
--- a/pyatlan/generator/class_generator.py
+++ b/pyatlan/generator/class_generator.py
@@ -99,8 +99,68 @@ TEMPLATES_DIR = PARENT / "templates"
 # applies when the named supertype is actually present in the typedef,
 # so a typedef change that drops it won't silently break.
 _SUPERCLASS_OVERRIDES: Dict[str, str] = {
+    # SHA-887: dbt-semantic types — Atlas typedef lists ``Dbt`` first but the
+    # publish-app ordering graph (and the asset's true type hierarchy) needs
+    # the semantic-family parent.
     "DbtMeasure": "SemanticMeasure",
+    # SHA-887 follow-up: dbt-process types — Atlas lists ``Dbt`` first but the
+    # publish-app's ``issubclass(cls, Process)`` / ``issubclass(cls,
+    # ColumnProcess)`` checks (used to identify "lineage types" that must
+    # publish *after* the entities they reference) need the process-family
+    # parent. Mis-ordering causes ATLAS-404 on dbt and SAP BW connectors.
+    "DbtProcess": "Process",
+    "DbtColumnProcess": "ColumnProcess",
 }
+
+# Per-module runtime injection overrides for typing.get_type_hints()
+# compatibility. Keyed by core asset module name; each value is the list of
+# class names whose imports are gated under ``if TYPE_CHECKING:`` in that
+# module's source (to break circular imports). After all assets load,
+# ``core/__init__.py`` writes these names back into each module's __dict__
+# so downstream tools (e.g. atlan-publish-app's ordering graph) can resolve
+# the field annotations without supplying a custom ``localns``. Keep this in
+# sync with the ``if TYPE_CHECKING:`` blocks in ``templates/module.jinja2``.
+_RUNTIME_INJECTION_OVERRIDES: Dict[str, List[str]] = {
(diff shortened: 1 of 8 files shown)


Aryamanz29:
Superseded — splitting into two cleaner PRs (atlan-python parent-class fixes + atlan-publish-app switch to pydantic ModelField). The get_type_hints shim added too much pyatlan-side maintenance burden; the better fix is publish-app moving off typing.get_type_hints entirely. New PRs incoming.

PR #932, opened 2026-05-14
fix: default daap_visibility on DataProduct.creator (BLDX-1252)

vaibhavatlan:
## Summary

- Default `daap_visibility = PRIVATE` in `DataProduct.creator(...)` / `DataProduct.Attributes.create(...)`, mirroring the Atlan UI default. Fixes the marketplace **Overview Assets** tile rendering blank for products created via the SDK.
- Add optional overrides on `creator()` / `create()` for `daap_visibility`, `daap_visibility_users`, `daap_visibility_groups`, `owner_users`, `owner_groups`.
- Update the matching `pyatlan/generator/templates/methods/{asset,attribute}/data_product.jinja2` templates so regeneration stays in sync.

Linear: https://linear.app/atlan-epd/issue/BLDX-1252

## Root cause

`DataProduct.Attributes.create(...)` previously set only `daap_status = ACTIVE` and left `daap_visibility` null. The marketplace Overview tile requires both `daapStatus` and `daapVisibility` to be non-null to fetch and render the Assets count, so SDK-created products appeared with a blank count even though `dataProductAssetsDSL` matched assets correctly. The earlier partial fix in PR #508 only addressed `daap_status`. Per @chris-grote's review on the Linear issue, the UI defaults Visibility to **Private**, so we mirror that here.

## Changes

- `pyatlan/model/assets/core/data_product.py`: extend `creator()` / `create()` / `Attributes.create()` with optional visibility / owners parameters; default `daap_visibility` to `DataProductVisibility.PRIVATE`.
- `pyatlan/generator/templates/methods/asset/data_product.jinja2` and `pyatlan/generator/templates/methods/attribute/data_product.jinja2`: same edits so generated output stays consistent.
- `tests/unit/model/data_product_test.py`: assert `daap_visibility == PRIVATE` on the default path; add `test_create_with_overrides` to cover the new params.

## Test plan

- [x] `pytest tests/unit/model/data_product_test.py` — all 9 tests pass
- [x] `ruff format --check` on changed Python files — clean
- [ ] Manual: create a DataProduct via `DataProduct.creator(...)` against a tenant and confirm the marketplace **Overview Assets** count renders (no longer blank)

### DIFF
--- a/pyatlan/generator/templates/methods/asset/data_product.jinja2
+++ b/pyatlan/generator/templates/methods/asset/data_product.jinja2
@@ -7,12 +7,19 @@
         name: StrictStr,
         domain_qualified_name: StrictStr,
         asset_selection: IndexSearchRequest,
+        client: Optional[AtlanClient] = None,
         daap_visibility: Optional[DataProductVisibility] = None,
         daap_visibility_users: Optional[Set[str]] = None,
         daap_visibility_groups: Optional[Set[str]] = None,
         owner_users: Optional[Set[str]] = None,
         owner_groups: Optional[Set[str]] = None,
     ) -> DataProduct:
+        # Mirror UI default: when a client is provided and no owners are
+        # specified, seed owner_users with the calling user.
+        if client is not None and owner_users is None:
+            current_username = client.user.get_current().username
+            if current_username:
+                owner_users = {current_username}
         attributes = DataProduct.Attributes.create(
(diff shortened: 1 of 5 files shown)


Aryamanz29 (commented):
LGTM - thanks!

Aryamanz29 (approved):

PR #950, opened 2026-06-09
feat: expose user publish preflight endpoint via workflow client (HYP-829)

rupeshatlan:
## Summary

Adds `client.workflow.user_publish_preflight(user_id, connection_qualified_name)` to the **v9 sync and async** workflow clients, wrapping Heracles' `POST /workflows/preflight/user-publish-check`.

The endpoint verifies the workflow creator's account is **enabled** and that they may **publish (ENTITY_CREATE/UPDATE/DELETE)** to the target connection. All checks — including the Keycloak impersonation/token-exchange — run **server-side in Heracles**, so callers only need their own service-account token; the endpoint is restricted to `service-account-*` callers on the Heracles side (`x-endpoint-type: service-account`).

## Why

Per @cmgrote's review on atlanhq/application-sdk#1966 ("New Heracles endpoints should be exposed via pyatlan, and then the existing pyatlan bits used for the actual interactions"): the application-sdk's `run_publish_preflight` Temporal activity — the first activity of every extract workflow (HYP-829) — currently calls this endpoint with raw httpx. With this wrapper the SDK switches to the pyatlan client and inherits its auth handling, observability headers, retries, and pooling.

Deliberately **not** done client-side via `client.impersonate.user` + `/evaluates`: that would require distributing impersonate-any-user credentials (`CLIENT_ID`/`CLIENT_SECRET`) to every connector app, including customer-deployed SDR runtimes. Keeping impersonation server-side in Heracles is the security boundary this design preserves (discussion: HYP-829).

## Changes (house pattern, +111/-0)

| File | Change |
|---|---|
| `pyatlan/client/constants.py` | `USER_PUBLISH_PREFLIGHT` API constant (`EndPoint.HERACLES`) |
| `pyatlan/client/common/workflow.py` | `WorkflowUserPublishPreflight` shared prepare/process |
| `pyatlan/client/common/__init__.py` | re-export |
| `pyatlan_v9/client/workflow.py` | sync `user_publish_preflight()` |
| `pyatlan_v9/client/aio/workflow.py` | async `user_publish_preflight()` |
| `tests/unit/test_workflow_client.py` | 3 unit tests for the shared logic |

## Related

- Heracles endpoint: atlanhq/heracles#5928
- SDK consumer: atlanhq/application-sdk#1966
- Linear: [HYP-829](https://linear.app/atlan-epd/issue/HYP-829)

## Notes for reviewers

- Files compile; unit tests are pure (no client mock needed) — but I could not execute the suite locally (local Python 3.14 can't import pyatlan's pydantic-v1 models), so relying on CI.
- v9-only by design (the SDK consumes `pyatlan_v9`); happy to mirror onto the legacy v1 `WorkflowClient` if wanted.
- Method returns the raw verdict `dict` (`{passed, failed_checks, message}`); can switch to a typed response model if preferred.

### DIFF
--- a/pyatlan/client/common/__init__.py
+++ b/pyatlan/client/common/__init__.py
@@ -192,6 +192,7 @@ from .workflow import (
     WorkflowStop,
     WorkflowUpdate,
     WorkflowUpdateOwner,
+    WorkflowUserPublishPreflight,
 )
 
 __all__ = [
@@ -337,4 +338,5 @@ __all__ = [
     "WorkflowStop",
     "WorkflowUpdate",
     "WorkflowUpdateOwner",
+    "WorkflowUserPublishPreflight",
 ]
--- a/pyatlan/client/common/workflow.py
+++ b/pyatlan/client/common/workflow.py
@@ -17,6 +17,7 @@ from pyatlan.client.constants import (
     SCHEDULE_QUERY_WORKFLOWS_MISSED,
     SCHEDULE_QUERY_WORKFLOWS_SEARCH,
     STOP_WORKFLOW_RUN,
+    USER_PUBLISH_PREFLIGHT,
     WORKFLOW_ARCHIVE,
     WORKFLOW_CHANGE_OWNER,
     WORKFLOW_INDEX_RUN_SEARCH,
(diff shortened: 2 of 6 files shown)


rupeshatlan:
Closing for now per HYP-829 final decision: the SDK will not call this endpoint — the Automation Engine calls Heracles directly (own client stack). Can be reopened if a pyatlan consumer appears.

PR #951, opened 2026-06-15
fix(batch): count partial updates and fix PARTIAL_UPDATE alias collision

aditya-atlan:
## Problem

`Batch`/`AsyncBatch._track_response` only inspect `mutated_entities.CREATE` and `.UPDATE`. But Atlan reports many **successful** updates as *partial*:

- primitive-only attribute changes (e.g. `user_description` alone) arrive under the top-level `partial_updated_entities` list
- any change touching a relationship (e.g. a glossary term's `anchor`) arrives under `mutated_entities.PARTIAL_UPDATE`

These were never counted. Worse, the sent-loop fallback bucketed them as `restored` with `num_updated = 0`. Any caller that gates on the update count therefore treats a successful update as a no-op — which, for consumers that retry on "nothing updated", causes a retry storm on the common glossary-term path.

There is also a latent model bug: `MutatedEntities.PARTIAL_UPDATE` was declared with `alias="DELETE"` (a copy-paste error). It collided with the real `DELETE` field, so a wire `"PARTIAL_UPDATE"` payload was dropped and deleted assets leaked into `assets_partially_updated()`. Only the top-level `partial_updated_entities` was usable before this fix.

## Changes

- **`model/response.py`**: fix `MutatedEntities.PARTIAL_UPDATE` alias (`"DELETE"` → `"PARTIAL_UPDATE"`).
- **`client/asset.py` (`Batch`) and `client/aio/batch.py` (`AsyncBatch`)**:
  - Reconcile partial updates from **both** `mutated_entities.PARTIAL_UPDATE` and the top-level `partial_updated_entities`, de-duplicated by GUID, into new `partial_updated` / `num_partial_updated` trackers.
  - Exclude partially-updated GUIDs from the `restored` fallback so they are no longer mislabeled.

## Behaviour

Same partial-update response, before vs after:

| | before | after |
|---|---|---|
| `num_updated` | 0 | 0 |
| `num_partial_updated` | n/a | 2 |
| `num_restored` | 2 (wrong) | 0 |

## Tests

- `test_partial_update_alias_does_not_collide_with_delete` — model deserialization keeps `DELETE` and `PARTIAL_UPDATE` distinct.
- `test_partial_updates_counted_and_not_restored` — sync + async: partial updates from both sources counted, de-duplicated, and not mislabeled as restored.

All existing `Batch`/`AsyncBatch` tests pass (391 in the client suites). `ruff check` / `ruff format` clean; `mypy` introduces no new errors.

[...]

PR #953, opened 2026-06-16
fix(batch): preserve real guid + qualified_name in Batch tracked lists

aditya-atlan:
## Problem

`Batch.__track` / `AsyncBatch.__track` (in **both** `pyatlan` and `pyatlan_v9`) rebuild each tracked asset via `trim_to_required()` / `ref_by_guid()` and then copy only `name`. So the tracked lists — `batch.created` / `updated` / `partial_updated` / `restored` — carry a **corrupted identity** that callers can't match back to their inputs:

- **non-glossary assets** (`trim_to_required()`): real `qualified_name` is kept, but the `guid` becomes a random placeholder (pyatlan, via `@init_guid`) or is left `UNSET` (pyatlan_v9).
- **glossary terms/categories** (`ref_by_guid()`): real `guid` is kept, but `qualified_name` is set **to the guid** (pyatlan) or left `UNSET` (pyatlan_v9).

Net: neither `guid` nor `qualified_name` is reliable across the tracked lists, so they can't be used to reconcile results across asset types. (This is distinct from the 9.7.5 partial-update count / alias fix in #951 — that was the count side; this is the read/identity side.)

## Fix

After building the tracked asset, restore the candidate's real `guid` and `qualified_name` (guarded for falsy / `UNSET`), keeping `name`. Applied to all four implementations:
- `pyatlan/client/asset.py` (`Batch`), `pyatlan/client/aio/batch.py` (`AsyncBatch`)
- `pyatlan_v9/client/asset.py` (`Batch`), `pyatlan_v9/client/aio/batch.py` (`AsyncBatch`)

## Verification

**Live (real calls against a tenant), with the fix:**
```
Table tracked → guid=44346641-...  qn=default/snowflake/.../ZZ_T          ✓ real
Term  tracked → guid=079055fa-...  qn=hqwJA39...@...  (qn != guid)        ✓ real
```

**Tests:** regression test added in all four suites (`tests/unit{,/aio}/test_client.py`, `tests_v9/unit{,/aio}/test_client.py`) asserting a tracked `Table` and `AtlasGlossaryTerm` each carry the real `guid` AND `qualified_name` (and a term's tracked `qualified_name` is no longer its guid). Confirmed failing pre-fix, passing post-fix. Full suites: **7583 passed, 5 skipped**. `ruff check`/`format` clean; `mypy` adds no new errors.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 
 ### Bug Fixes
 
-- **`Batch`/`AsyncBatch` tracked lists now carry real identity (both `pyatlan` and `pyatlan_v9`)**: the tracked lists (`created` / `updated` / `partial_updated` / `restored`) now carry each asset's real `guid` and `qualified_name`, so callers can match tracked items back to their inputs by either key. Previously non-glossary assets got a placeholder (negative) guid in `pyatlan` (or a missing/`UNSET` guid in `pyatlan_v9`) via `trim_to_required()`, and glossary terms got the guid in place of `qualified_name` in `pyatlan` (or a missing/`UNSET` `qualified_name` in `pyatlan_v9`) via `ref_by_guid()`, making the tracked items impossible to match back by guid or qualified_name.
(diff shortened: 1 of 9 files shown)


Aryamanz29 (approved):
LGTM

PR #971, opened 2026-07-13
fix: clear NotFoundError on tag-cache miss instead of (DELETED) sentinel (BLDX-1530)

Aryamanz29:
## What & why

Fixes **BLDX-1530** (Multiplex Sev1, ZD#126376). When a tag name can't be resolved in the local tag cache, `AtlanTagRetranslator` substituted the literal `(DELETED)` sentinel into the request payload and sent it to Atlas, which rejected it with an opaque:

```
ATLAS-404-00-008: Given classification (DELETED) was invalid.
```

That names neither the tag nor the cause, and misleads customers into thinking a tag was *deleted* when it may simply **never have existed** — the repro is `add_atlan_tags(["Alert: DQ"])` on a tenant without that tag.

## The fix

On the **asset-tag write path** (classification objects: `classifications` / `addOrUpdateClassifications` / `removeClassifications`), an unresolvable tag name now raises a clear client-side error naming the tag:

```
NotFoundError: ATLAN-PYTHON-404-006 Atlan tag with name Alert: DQ does not exist.
```

Two behaviors are deliberately preserved:
- **`(DELETED)` sentinel round-trip** — an asset *read* with an already-deleted tag surfaces `(DELETED)`; re-serializing it must stay lossless, so a name that *is* the sentinel is preserved (not raised on).
- **purpose/persona policy (names) path** — `classificationNames` / `purposeClassifications` stay tolerant (they round-trip deleted references through the sentinel); the strict check applies only to the asset-tag write path where the customer-facing bug occurs.

Fixed in **both** the sync (`retranslators.py`) and async (`aio/retranslators.py`) retranslators.

## Tests
- **Unit** (`tests/unit/test_retranslators.py` + `aio/`): resolve existing tag → id; missing tag → named `NotFoundError` (no sentinel leak); `removeClassifications` missing → raises; `(DELETED)` sentinel round-trip preserved; names/policy path stays tolerant.
- **Integration** (`tests/integration/test_client.py::test_add_nonexistent_classification_raises_named_error`): `add_atlan_tags` with a non-existent tag raises the named error end-to-end.
- Full unit suite: **6992 passed**, no regression.

Related: CSA-491 (the customer package crash), DOC-815 (same opaque symptom for playbooks).

Closes BLDX-1530.


Aryamanz29:
### Update: live testing changed the approach

Validated the earlier `deleted_names`-based distinction on a real tenant (kill-argo) and found it **never worked**, so I've corrected it.

**Root cause of the miss:** `AtlanTagCacheCommon.get_id_for_name_after_refresh` adds *any* unresolved name to `deleted_names` on a refresh miss (comment: "could be an audit-log entry"). Confirmed live — after one lookup, a never-existed name is already `in deleted_names`. So `deleted_names` conflates *soft-deleted* with *never-existed*; gating the raise on it meant the guard never fired, and `add_atlan_tags(["<missing>"])` still shipped `(DELETED)` to Atlas → the original opaque `ATLAS-404-00-008: Given classification (DELETED) was invalid`.

**Fix:** the cache can't tell the two apart by name, so the only reliable "genuinely deleted" signal is the `(DELETED)` **sentinel string** itself (produced by a *read* of a deleted tag). Gate on that: preserve the sentinel (lossless round-trip), raise a clear named `NotFoundError` (`ATLAN-PYTHON-404-006`) for anything else unresolvable.

**Live validation (all three cases, on a real tenant):**
| Case | Result |
|---|---|
| `add_atlan_tags(["NonExistent-…"])` | ✅ `ATLAN-PYTHON-404-006 Atlan tag with name NonExistent-… does not exist` (named, no sentinel leak) |
| `add_atlan_tags(["<active tag>"])` | ✅ succeeds |
| read asset whose tag was deleted | ✅ `(DELETED)` sentinel preserved |

Unit test now pins that `deleted_names` membership does **not** suppress the raise (regression guard for this exact bug); integration test asserts the specific `404-006` message and absence of `(DELETED)`.

PR #1046, opened 2026-09-16
feat(models): regenerate pydantic assets from latest typedefs

Aryamanz29:
## What

Regenerates the pydantic `pyatlan/model/assets` (+ `structs`, `enums`, docs) from the latest Atlan typedefs, bringing the pydantic SDK in line with the models-master typedefs (the pydantic counterpart to the pyatlan_v9 sync in #1042). Contains two commits:

1. **`fix(generator)`** — a generator bug fix (details below), required for the regen to import.
2. **`feat(models)`** — the regenerated assets.

## Source & tenant validation

The class generator seeds from a live tenant's typedefs. I validated three tenants against the authoritative Pkl-sourced `pyatlan_v9` models before choosing one:

| Tenant | vs models-master (v9) | Verdict |
|---|---|---|
| eng tenant A | missing ~400 attributes (e.g. `skillArtifactContent`, `sqlCoalesce*`) | stale |
| eng tenant B | retains 20 deprecated types (Coalesce/Cortex/KafkaPartition/…) | stale |
| **eng tenant C (used)** | deprecated types gone, all attributes present | **aligned** |

Generated from the aligned tenant. Cross-checked every dropped attribute against `pyatlan_v9`: **no authoritative attribute is dropped** — the only removed getters are relationships to genuinely-deleted types.

## Changes

- **New:** `DatabricksGenieAgent`; `KnowledgeFile` / `KnowledgeFolder` (and their `Knowledge` base) promoted into `core/`.
- **Removed (platform-deprecated):** `Coalesce`/`V1Coalesce` family, Snowflake V1 Cortex types, `KafkaPartition`, `SigmaDataModelElement`, and their relationship attributes.

## Generator fix (commit 1)

The generator promotes a core asset's *referenced* assets to core to a fixpoint, but not its *super-class* chain. `KnowledgeFolder` gets pulled into `core/` as a referenced asset of `Asset`, but its parent `Knowledge` was left non-core. A `core/` module imports its super-class eagerly at the top of the file, so a non-core parent gets pulled across the `assets/ → assets/core/` boundary mid-cycle and its `update_forward_refs()` runs before the namespace is complete — failing at import with `NameError: name 'AtlasGlossaryTerm' is not defined`. The fix extends the existing fixpoint closure to also promote each core asset's super-class, so core is closed under super-class (promote-only, order-independent). Net effect on this regen: exactly one extra file (`core/knowledge.py`) joins core.

## Validation

- `ruff format --check` + `ruff check --no-fix`: clean
- `mypy pyatlan`: clean
- `tests/unit`: **7108 passed, 0 failed, 5 skipped**
- Runtime import of the full asset package + new/moved types: OK


pyatlan/model/assets/core/__init__.py
```diff
@@ -8,209 +8,610 @@
...
+
+
+from .agentic import Agentic
+
+
+from .b_i import BI
+
+
+from .semantic import Semantic
+
```
Aryamanz29: so much whitespace

